PR #12, opened 2015-07-09
update README to include creating the ApplicationCabel::Channel

lsylvester:
I was getting `uninitialized constant ApplicationCable::Channel`.

I have updated the documentation to include creating the `ApplicationCable::Channel` class.

However, I am not sure of the point of this class instead of directly subclassing `ActionCable::Cannel::Base`.

Also, there was reference to `ApplicationChannel` whichI  have  was replaced with `ApplicationCable::Channel` for consistency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,18 @@ module ApplicationCable
 end
 ```
 
+Then you should define your `ApplicationCable::Channel` class in Ruby. This is the place where you put
+shared logic between your channels.
+
+```ruby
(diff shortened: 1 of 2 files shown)


dhh:
Thanks! We've been moving to a setup of having an ApplicationX super-class for all frameworks, such that the app can have an obvious place for app-configuration that then won't conflict with stuff added via engines or plugins.

PR #17, opened 2015-07-09
Add channel generator

seuros:
cc/ @dhh

### DIFF
--- a/lib/rails/generators/channel/channel_generator.rb
+++ b/lib/rails/generators/channel/channel_generator.rb
@@ -1,0 +1,26 @@
+require 'rails/generators/named_base'
+
+module Rails # :nodoc:
+  module Generators # :nodoc:
+    class ChannelGenerator < Rails::Generators::NamedBase # :nodoc:
+      desc 'This generator creates a cable channel file in app/channels'
+      source_root File.expand_path('../templates', __FILE__)
+      check_class_collision suffix: 'Channel'
+
+      hook_for :test_framework
+
+      def create_channel_files
+        template 'channel.rb.erb', File.join('app/channels', class_path, "#{file_name}_channel.rb")
+        template 'channel.coffee.erb', File.join('app/assets/javascripts/channels', "#{file_name}_channel.coffee")
+      end
+
+      def channel_name
+        lass_name.camelize(:lower)
+      end
+
+      def channel_class
+        "#{class_name}Channel"
+      end
+    end
+  end
+end
--- a/lib/rails/generators/channel/templates/channel.coffee.erb
+++ b/lib/rails/generators/channel/templates/channel.coffee.erb
@@ -1,0 +1,6 @@
+@App ?= {}
+App.<%= channel_name %> = App.cable.subscriptions.create "<%= channel_class %>",
+  connected: ->
+    # Called once the subscription has been successfully completed
+  received: (data) ->
+    # Called when data is received
--- a/lib/rails/generators/channel/templates/channel.rb.erb
+++ b/lib/rails/generators/channel/templates/channel.rb.erb
@@ -1,0 +1,11 @@
+<% module_namespacing do -%>
+class <%= channel_class %> < ApplicationCable::Channel
+  def subscribed
+    # Do something
+  end
+
+  def unsubscribed
+    # Do something
+  end
+end
+<% end -%>


lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      def channel_name
+        class_name.underscore
+      end
+
+      def channel_class
+        "#{class_name}Channel"
+      end
+
+      def parent_class_name
```
dhh: Don't think we need this. It should just always be ApplicationCable::Channel. Just like the controller generator always generates controllers inheriting from ApplicationController.

lib/rails/generators/channel/templates/channel.coffee.erb
```diff
@@ -0,0 +1,3 @@
+App.<%= channel_name %> = App.cable.subscriptions.create "<%= channel_class %>",
+  connected: ->
+    # Called once the subscription has been successfully completed
```
dhh: Should add an example `received: (data) ->` method as well.

lib/rails/generators/channel/templates/channel.coffee.erb
```diff
@@ -0,0 +1,3 @@
+App.<%= channel_name %> = App.cable.subscriptions.create "<%= channel_class %>",
```
dhh: We need to declare `@App = {}` somewhere for this to work. But I guess that can come with the integration with Rails as well, where we can add that to application.js.

seuros: What about ```App ||= {}```  for now ?

dhh: Yeah, probably fine for now. Gotta be `@App`, though, as it needs to be
exported.

On Thu, Jul 9, 2015 at 11:41 AM, Abdelkader Boudih <notifications@github.com

seuros: Done

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      def channel_name
+        class_name.underscore
+      end
+
+      def channel_class
+        "#{class_name}Channel"
+      end
+
```
dhh: :scissors: extra line

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails/generators/named_base'
+
+module Rails # :nodoc:
+  module Generators # :nodoc:
+    class ChannelGenerator < Rails::Generators::NamedBase # :nodoc:
+      desc 'This generator creates a cable channel file at app/channels'
```
dhh: at -> in

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,29 @@
...
+      source_root File.expand_path('../templates', __FILE__)
+      check_class_collision suffix: 'Channel'
+
+      class_option :parent, :type => :string, :desc => 'The parent class for the generated channel'
```
dhh: Use 1.9 hash syntax.

dhh: Actually, sorry, we should kill this whole parent option.

lib/rails/generators/channel/templates/channel.coffee.erb
dhh: Isn't it `?=` in CS?

seuros: both works. I will use ```?=```

javan: This will generate JavaScript that creates an error unless the user has already defined `App.cable = Cable.createConsumer(...)`, and redeclaring a global `@App` in every file isn't ideal for those who don't wish to use that namespace.

Perhaps we should standardize defining your consumer and subscriptions on `@Cable` itself. Something like:

```coffee
@Cable =  
  consumer: null
  subscriptions: {}
```

And the generator could generate something like:
```coffee
Cable.subscriptions.<%= channel_name %> = Cable.consumer?.subscriptions.create "<%= channel_class %>", ...
```

dhh: I like that, Javan. Depends on having a cable JS initializer, but I think
we should have that anyway. +1 on putting this on Cable rather than App
too. Let's get that in and flow that through all the examples etc.

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      hook_for :test_framework
+
+      def create_channel_files
+        template 'channel.rb.erb', File.join('app/channels', class_path, "#{file_name}_channel.rb")
+        template 'channel.coffee.erb', File.join('app/assets/javascripts/channels', "#{file_name}_channel.coffee")
+      end
+
+      def channel_name
+        class_name.underscore
```
javan: If this is going to be used as a JavaScript variable, it should be camelCased as `class_name.camelize(:lower)`.

lib/rails/generators/channel/templates/channel.coffee.erb
seuros: Should we support the .js format aswell ? In case CS is not part of the bundle.

javan: All of Cable's JavaScript is written in CS so it's fine to generate .coffee for now.

seuros: @javan 
What do you think about: 
```
Cable.channels.<%= channel_name %> = Cable.subscriptions.create "<%= channel_class %>", ...

javan:
Thinking on this more, I don't think we should be generating client-side "channels" automatically. For one, they're subscriptions, not channels. There isn't necessarily a one-to-one mapping between channels on the server and subscriptions on the client. You can have multiple subscriptions to a channel, and you can also subscribe to the same channel, but with different parameters: `App.cable.subscriptions.create("ChatChannel")` and `App.cable.subscriptions.create({channel: "ChatChannel", room: 123 })`.

Even if we did store subscriptions in `Cable.subscriptions`, they still need to be created from a `Cable.Consumer` instance and there's no guarantee one will be present. So we have to code defensively with `Cable.consumer?.subscriptions.create`. Auto-generated templates like this should Just Work™ and these might not.

The generator for server channels makes perfect sense so :+1: on that. Not sure what's possible, but perhaps the generator could print example subscription JS to stdout. And ultimately I think we should rely on a good JS documentation for client subscriptions.

dhh:
The generator is mainly there as instructive code. Teaching people how to
use this. I think it's fine that it generates a stub you may well want to
change, and I also think it's fine that they expect that some setup has
already been done.

We should make it easy to do this setup by shipping a default JS
configuration for the cable that's commented out, which people can comment
in to make it work.

rafaelfranca:
Thank you for the pull request. It is already integrated

PR #19, opened 2015-07-09
Update readme instructions for js

lsylvester:
Update javascript README so that:

* The actioncable js is required
* The filename for example js doesn't conflict with the actioncable js


javan:
Since _all_ of the other examples are .coffee files, why don't we stick with it and combine these into one example?

How about:

```coffee
# app/assets/javascripts/application_cable.coffee

#= require cable

 @App = {}
 App.cable = Cable.createConsumer "ws://cable.example.com"
```

PR #20, opened 2015-07-10
Replace assignment with equality check in example

wurde:
First time reading the ActionCable README, but I believe this ought to be an equality check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,14 @@ module ApplicationCable
     end
 
     protected
-      def find_verified_user
-        if current_user = User.find(cookies.signed[:user_id])
-          current_user
-        else
-          reject_unauthorized_connection
-        end
+
+    def find_verified_user
+      if current_user == User.find(cookies.signed[:user_id])
+        current_user
+      else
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,13 +51,14 @@ module ApplicationCable
...
     protected
-      def find_verified_user
-        if current_user = User.find(cookies.signed[:user_id])
-          current_user
-        else
-          reject_unauthorized_connection
-        end
+
+    def find_verified_user
+      if current_user == User.find(cookies.signed[:user_id])
```
seuros: ```User.find_by_id(cookies.signed[:user_id])``` since ```#find``` will raise if the user don't exists.

lsylvester:
I think that it should be an assignment, otherwise the `current_user` that is returned on the next line will be nil, and then `self.current_user` will be set to nil.

wurde:
But Rubocop didn't like assignments in conditions. I think you're right though and I'll take a closer look.

wurde:
Confirmed, thank you.

PR #22, opened 2015-07-10
Initial work on getting some tests.

lsylvester:
These changes are to get the tests running and add a sample test to show how the server can be tested.

I was not sure how to tell when it would be ok to broadcast a message - so it currently waits for the first ping message, which makes the test a bit slow, so feedback on the best way to do that would be good.


IlyaBylich:
Asynchronous stuff should be tested in a different way. For testing the following code:
``` ruby
def some_method(&callback)
  callback.call
end
```
we should not just right something like
``` ruby
test 'test message' do
  some_method do
    assert true
  end
end
```
The code below doesn't test the case when your method doesn't actually invoke passed block. Probably the interface of testing framework (`ActionCableTest`) should be changed to support async tests.

For example, jasmine takes an argument called `done` (a callback) to the test definition. And when you write jasmine spec, you must manually invoke this callback when all assertions were made. If you don't call it - the test becomes red.

To make this working with `ActiveSupport::TestCase` (I think) we need something like https://gist.github.com/IlyaBylich/50bcd1ee4826ab9b8a17 (this example, actually works, but as for me, it's super ugly)

Any other ideas about it?

lsylvester:
It is only the code inside the EM.run block that is asynchronous, not the whole test. The EM.run block will block until  EM.stop_event_loop is called, so in effect EM.stop_event_loop is like done callback in the jasmine test.

lsylvester:
Looks likes the tests are working now so this is not required.

PR #23, opened 2015-07-10
Ensure ActionCable is thread-safe

rosenfeld:
Although this is unlikely to cause any problems specially because this
is first run as "run ActionCable.server" and it's very unlikely to call
multiple rack "run" methods concurrently and even if we did so it's not
guaranteed that would cause any troubles, it's much better not having to
worry about this and just ensuring this is thread-safe.

It might seem that @a ||= A.new pattern runs atomically, but this is not
guaranteed nor even by MRI. Here's an example:

```ruby
class A
  def initialize
      sleep 1
  end
  def self.instance
     @@instance ||= new
  end
end

instances = [] # I know this is not thread-safe but it demonstrates the problem anyway

(1..2).map{Thread.start{instances << A.instance}}.each &:join
instances[0] == instances[1] # false

Using the singleton pattern makes the intention clearer and worry-free.

### DIFF
--- a/lib/action_cable.rb
+++ b/lib/action_cable.rb
@@ -26,8 +26,7 @@ module ActionCable
   autoload :RemoteConnections, 'action_cable/remote_connections'
   autoload :Broadcaster, 'action_cable/broadcaster'
 
-  # Singleton instance of the server
   module_function def server
-    @server ||= ActionCable::Server::Base.new
+    ActionCable::Server::Base.instance
   end
 end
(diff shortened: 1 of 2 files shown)


seuros:
LGTM

rosenfeld:
I'm also curious on why `module_function def server` rather than simply `def self.server`. Is there any intentions in mixin in this module somewhere?

lifo:
In a distant future, we'd like to allow running multiple cable servers for the same Rails app. Some could be run as a middleware. Some could be a part of an engine.

In any case, I don't think this change would have any real benefits until then.

PR #25, opened 2015-07-13
 Update README to use coffeescript  all the time

lsylvester:
Remove the js in the README as per @javan's feedback on #19.


robin850:
Thanks! :+1:

PR #28, opened 2015-07-14
[PoC] add rabbitmq as an adapter for actioncable

meinac:
I don't want to waste your time, this is just a poc work. Maybe we can add other message brokers like ActiveMQ or Kafka or maybe we can add AWS SNS I believe we can use it for pubsub operations and also it's fully managed service so as a developer I don't need to manage the server. Or maybe somebody will create an external gem as an adapter for these but at first we need to separate Redis from source of the actioncable as an adapter like ActiveRecord's connection adapters and we need to create interface which other adapters can implement it. So what do you think about these ideas?

### DIFF
--- a/actioncable.gemspec
+++ b/actioncable.gemspec
@@ -18,6 +18,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'celluloid',        '~> 0.16.0'
   s.add_dependency 'em-hiredis',       '~> 0.3.0'
   s.add_dependency 'redis',            '~> 3.0'
+  s.add_dependency 'amqp'
 
   s.files = Dir['README', 'lib/**/*']
   s.has_rdoc = false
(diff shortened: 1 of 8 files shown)


guilleiguaran:
@meinac hey!! I think your idea is good but let's start first moving the redis adapter to a separated class in a separate PR and then we can discuss later if we want to add more adapters to the core gem or if this should be done in a separated gem, either way I think moving the Redis dependency to an adapter class is a good idea to support other adapters.

dhh:
Concur with @guilleiguaran. We're not going to want to depend on any other adapters than Redis as the default, but would be fine to have a adapter system that allows the user to supply the adapters.

Another use case for this I've found is to use r/w splitting Redis adapters when you run something in a multi-dc setup. So we'd use it just for that.

meinac:
@guilleiguaran , @dhh agreed. I'm currently working on this.

PR #29, opened 2015-07-17
create adapter system

meinac:
The idea comes from #28 
/cc @guilleiguaran , @dhh

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,15 +206,19 @@ See the [rails/actioncable-examples](http://github.com/rails/actioncable-example
 ## Configuration
 
 The only must-configure part of Action Cable is the Redis connection. By default, `ActionCable::Server::Base` will look for a configuration
-file in `Rails.root.join('config/redis/cable.yml')`. The file must follow the following format:
+file in `Rails.root.join('config/actioncable.yml')`. The file must follow the following format:
 
 ```yaml
+default: &default
+  :adapter: redis
 production: &production
+  <<: *default
   :url: redis://10.10.3.153:6381
   :host: 10.10.3.153
   :port: 6381
   :timeout: 1
 development: &development
+  <<: *default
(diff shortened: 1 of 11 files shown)


matthewd:
There seems to be too many layers of indirection here.

`Broadcaster` contains a `@broadcaster`, which it populates by calling `Redis::Broadcaster.new.broadcaster` ? :confused:

dhh:
Yeah, appreciate the initial work here, but this isn't quite right. Going to take a stab at this shortly.

cristianbica:
@dhh do you want something like [this](https://github.com/rails/actioncable/compare/rails:master...cristianbica:adapter)?
Then we would have a `ActionCable::Base.connection` which will be an instance of an adapter and it will be used across classes.
I would also like to have an AR style configuration `config/action_cable.yml` or `ENV['ACTION_CABLE_BACKEND_URL']`

inopinatus:
Hoping we can incorporate postgresql's native pub/sub as one such adapter. Minimising the external infrastructure dependencies will be a win for lean startups.

rafaelfranca:
Closing since it is not the direction that we want to take.

meinac:
@rafaelfranca, in the current implementation of this feature it depends on redis, but as @inopinatus mentioned as a developer if I want to use postgre as pubsub agent there is no solution for this. In this pr I wanted to separate pubsub agent like an activerecord adapter to give direction to open source developers to implement their third party adapters so they can implement adapters like aws sqs, postgresql, rabbitmq or whatever they like. This pr does not aim to implement another agent in itself, it's aim is only create a way for developers to implement their adapters.
By this pr, opensource developers can create their own adapters for their favorite pubsub agent and I think this is important to give developers to use existing infrastructure as pubsub agent.

PR #30, opened 2015-07-18
Server::Base broadcasting test added

Axxiss:
Hi, this is my first contribution attempt to rails community.

I created a simple test to check if a message is delivered to `Redis`, I used `Minitest::Mock` to isolate the server. Not quite sure if `ActionCable::Server::Configuration` should be mocked.

I would like to have some feedback, so I can move forward to add more tests.

### DIFF
--- a/test/server/base_test.rb
+++ b/test/server/base_test.rb
@@ -1,0 +1,32 @@
+require 'test_helper'
+
+class ActionCable::Server::BaseTest < ActiveSupport::TestCase
+
+  setup do
+    mock_config
+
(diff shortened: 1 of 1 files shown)


dhh:
Hey, welcome, and thanks for the contribution!

We shouldn't need to mock the config. That one can be created on its own and assigned to `ActionCable::Base.config =`.

dhh:
Let's stub Rails. Action Cable is not intended to be useful outside the context of Rails at the moment.

Axxiss:
New PR #38  opened using a feature branch.

PR #31, opened 2015-07-19
Add coffee-rails as a dependency.

constXife:


seuros:
CS is not a dependency of ACable. You can use plain js if you wish.

matthewd:
Do we need to be so particular about the version? [Turbolinks, for example,](https://github.com/rails/turbolinks/blob/master/turbolinks.gemspec#L14) leaves the versioning up to coffee-rails itself.

constXife:
@seuros, i'd like to. But in actioncable/lib/assets/javascripts I see only coffee files. 
@matthewd, you're right.

seuros:
@constXife yes, but CS is not a dependency. You can rename and use js or anything that generate javascript in the end.

rafaelfranca:
> yes, but CS is not a dependency.

Yes, it is. Users can't use this gem's  JavaScript code without coffee-script.

Issue #26, opened 2015-07-14
Generation of stream names RFC


lsylvester:
Currently, to stream events in a channel, you call steam_from with a string. This same string must be generated elsewhere in the application to broadcast events to this channel.

Instead of passing in a string, it could be possible to pass in `ActiveModel` compatible object and have the stream name generated from the model, and prefixed by the channel name.

So instead of:

```ruby
class WebNotificationsChannel < ApplicationCable::Channel
  stream_from "web_notifications_#{current_user.id}" 
end
```
You could have
```ruby
class WebNotificationsChannel < ApplicationCable::Channel
  stream_from current_user #=> streams from web_notifications:user_1
end
```

And then when broadcasting, instead of 
```ruby
ActionCable.server.broadcast \
  "web_notifications_1", { title: 'New things!', body: 'All shit fit for print' }
```

you could have
```ruby
WebNotificationsChannel.broadcast user, { title: 'New things!', body: 'All shit fit for print' }
  #=> broadcasts to web_notifications:user_1
```

I would be happy to make a PR for this if anyone thinks it is a good idea.

IlyaBylich:
Sounds amazing.
But probably it would be better to have some kind of concern (`ActionCable::Support` or something like this) that will be included into `ActiveModel::Model` by default and implements some common methods like `to_actioncable_channel`? It allows to mix this module into any class (that represents a business model, but is not fully compatible with `ActiveModel::Model`) and implement this method manually.

dhh:
I like this a lot. I'd propose a few changes. 

First, let's use the GlobalID for this. That's already a global unique identifier for a given object. We can use #to_gid_param for the identifier itself and the full identification for the logging.

Second, here's a twist to the APIs when using this type of composition:

```ruby
class WebNotificationsChannel < ApplicationCable::Channel
  def subscribed
    stream_for current_user #=> streams_from("web_notifications:Z2lkOi8vYmMzL1BlcnNvbi8x")
  end
end

WebNotificationsChannel.broadcast_to user, { title: 'New things!', body: 'All shit fit for print' }
# => WebNotificationsChannel.broadcast "web_notifications:Z2lkOi8vYmMzL1BlcnNvbi8x", ...
```

Then it's clear when we're deriving the broadcasting name vs when we're accepting the full one.

lsylvester:
[...]

PR #34, opened 2015-07-22
add broadcast_to and stream_for methods as per #26

lsylvester:


lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
+module ActionCable
+  module Channel
+    module Broadcasting
+
```
dhh: :scissors: newline.

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
...
+      # Broadcast a hash directly to a named <tt>broadcasting</tt>. It'll automatically be JSON encoded.
+      def broadcast(broadcasting, message)
```
dhh: I don't think this is really needed? It doesn't add anything over ActionCable.server.broadcast. Let's nix it.

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
...
+        ActionCable.server.broadcast(broadcasting, message)
+      end
+
+      # Broadcast a hash a unique broadcasting for this <tt>object</tt> in this channel.
+      def broadcast_to(object, message)
+        broadcast(broadcasting_for([channel_name, object]), message)
+      end
+
+      # Generates a unique named broadcasting for the <tt>object</th>.
+      def broadcasting_for(object)
```
dhh: Should be private.

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
...
+        case
+        when object.is_a?(Array)               then object.map{ |o| broadcasting_for(o) }.join(':')
```
dhh: Not loving the massive spacing needed for `then`. Let's go with newlines. Also, let's go with `model` as the var name over `object`, since we require `to_param` to be present (thus inferring we need active models).

Finally, space between map and the opening {

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
...
+        when object.respond_to?(:to_global_id) then object.to_global_id.to_param
```
dhh: We can use `to_gid_param`.

lib/action_cable/channel/streams.rb
```diff
@@ -63,6 +63,13 @@ def stream_from(broadcasting, callback = nil)
         logger.info "#{self.class.name} is streaming from #{broadcasting}"
       end
 
+      # Start streaming the pubsub queue for the <tt>object</tt> in this channel. Optionally, you can pass a
+      # <tt>callback</tt> that'll be used instead of the default of just transmitting the updates straight
+      # to the subscriber.
+      def stream_for(object,callback = nil)
```
dhh: Space after comma.

lib/action_cable/channel/streams.rb
```diff
@@ -63,6 +63,13 @@ def stream_from(broadcasting, callback = nil)
...
+        stream_from(self.class.broadcasting_for([channel_name,object]),callback)
```
dhh: `[channel_name,object]` => `[ channel_name, object ]`

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,23 @@
+module ActionCable
+  module Channel
+    module Naming
+      extend ActiveSupport::Concern
+
+      class_methods do
```
kaspth: Would be easier to go with:

```ruby
included do
  # Your documentation here...
  mattr_reader(:channel_name) { name.sub(/Channel$/, '').gsub('::',':').underscore }
end
```

Gives us the instance reader free too.

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,23 @@
...
+        # Returns the last part of the channels's name, underscored, without the ending
+        # <tt>Channel</tt>. For instance, ChatChannel returns <tt>chat</tt>.
+        #
+        # ==== Returns
+        # * <tt>string</tt>
```
kaspth: Let's make the example more real instead of this returns thing. Like so:

```ruby
# Returns the last part of the channels's name, underscored, without the <tt>Channel</tt> ending.
#
#   ChatChannel.channel_name # => 'chat'
```

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,23 @@
...
+        end
+      end
+
+      # Delegates to the class' <tt>controller_name</tt>
+      def channel_name
+        self.class.channel_name
+      end
+    end
+  end
+end
```
kaspth: Remember to put a new line at the end of the files you add.

lib/action_cable/channel/naming.rb
kaspth: Honestly, I'm not sure what you mean by `the last part` because it seems the channel name is scoped after the module namespace (we should also show an example of that).

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,25 @@
...
+        else                                        object.to_param
+        end.to_s
```
kaspth: Doesn't `to_param` return a string already? `join` definitely does, so I'd rather :scissors: it.

lib/action_cable/channel/base.rb
```diff
@@ -68,6 +68,8 @@ class Base
       include Callbacks
       include PeriodicTimers
       include Streams
+      include Naming
+      extend Broadcasting
```
kaspth: Doesn't make sense to have it at the class scope to me. Why do we need it?

test/channel/stream_test.rb
```diff
@@ -15,10 +17,15 @@ def subscribed
   end
 
   test "streaming start and stop" do
-    @connection.expects(:pubsub).returns mock().tap { |m| m.expects(:subscribe) }
+    @connection.expects(:pubsub).returns mock().tap { |m| m.expects(:subscribe).with("test_room_1") }
     channel = ChatChannel.new @connection, "{id: 1}", { id: 1 }
-
```
kaspth: New lines are great. Keep it in there :smile:

test/channel/broadcasting_test.rb
```diff
@@ -0,0 +1,34 @@
...
+class ActionCable::Channel::BroadcastingTest < ActiveSupport::TestCase
+  class ChatChannel < ActionCable::Channel::Base
+  end
+
+  setup do
+    @connection = TestConnection.new
+  end
+
+  test "broadcasting a message" do
+    ActionCable.stubs(:server).returns mock().tap{ |m| m.expects(:broadcast).with("test_room_1", "Hello World")}
```
kaspth: We also want a space between `tap` and the open brace.

lib/action_cable/channel/streams.rb
```diff
@@ -12,7 +12,7 @@ module Channel
     #     def follow(data)
     #       stream_from "comments_for_#{data['recording_id']}"
```
kaspth: Shouldn't we update the existing documentation to use `stream_for`?

lib/action_cable/channel/naming.rb
lsylvester: I tried this but the block was evaluated in `ActionCable::Channel::Base` instead of each of the subclasses so it was not generating the correct name.

kaspth: Ah, right. Nice catch :+1:

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,23 @@
...
+        # <tt>Channel</tt>. For instance, ChatChannel returns <tt>chat</tt>.
+        #
+        # ==== Returns
+        # * <tt>string</tt>
+        def channel_name
+          @channel_name ||= name.sub(/Channel$/, '').gsub('::',':').underscore
+        end
+      end
+
+      # Delegates to the class' <tt>controller_name</tt>
```
kaspth: How about `delegate :channel_name, to: :class` instead?

lib/action_cable/channel/broadcasting.rb
matthewd: I think this block would feel more natural spelled as an `if`.

lib/action_cable/channel/base.rb
lsylvester: The `Broadcasting` methods are at the class scope because we don't have any instance to the channel when broadcasting, and we want to be able to generate the broadcasting name based on the channels name.

Doers that make sense?

lib/action_cable/channel/naming.rb
lsylvester: @kaspath - You are right - `the last part` should not be there, and has been removed.

lib/action_cable/channel/base.rb
kaspth: Sounds to me like that's something the broadcasting concern should handle, so we don't have to say `self.class.broadcasting_for`.

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,21 @@
+module ActionCable
+  module Channel
+    module Broadcasting
+      # Broadcast a hash a unique broadcasting for this <tt>object</tt> in this channel.
```
kaspth: `a hash a unique broadcasting` - what does this mean?

Speaking more generally, I find the above documentation tough to follow.

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,21 @@
...
+      def broadcast_to(object, message)
```
kaspth: `object` => `model`.

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,21 @@
+module ActionCable
+  module Channel
+    module Naming
+      extend ActiveSupport::Concern
+
+      class_methods do
+        # Returns the channels's name, underscored, without the <tt>Channel</tt> ending.
+        #
+        #   ChatChannel.channel_name # => 'chat'
```
kaspth: We still need a namespaced example too:

```ruby
#   Chats::AppearancesChannel.channel_name # => 'chats:appearances'
```

lib/action_cable/channel/naming.rb
```diff
@@ -0,0 +1,21 @@
+module ActionCable
+  module Channel
+    module Naming
+      extend ActiveSupport::Concern
+
+      class_methods do
+        # Returns the channels's name, underscored, without the <tt>Channel</tt> ending.
```
kaspth: We should clarify that module namespaces are separated with colons.

(Also `channels's name` could be `the name of the channel`.)

lib/action_cable/channel/broadcasting.rb
lsylvester: I had missed a word, should have been "a hash to a unique broadcasting". I have been updated it.

lib/action_cable/channel/streams.rb
lsylvester: I have updated the second example in the stream docs to use `stream_for`. Is that what you were referring to, or do you think there needs to be an example dedicated to `stream_for`?

lib/action_cable/channel/streams.rb
```diff
@@ -63,6 +63,13 @@ def stream_from(broadcasting, callback = nil)
         logger.info "#{self.class.name} is streaming from #{broadcasting}"
       end
 
+      # Start streaming the pubsub queue for the <tt>model</tt> in this channel. Optionally, you can pass a
+      # <tt>callback</tt> that'll be used instead of the default of just transmitting the updates straight
+      # to the subscriber.
+      def stream_for(model, callback = nil)
+        stream_from(broadcasting_for([ channel_name, model ]),callback)
```
kaspth: We need a space between the comma and callback here too :smile:

test/channel/broadcasting_test.rb
```diff
@@ -0,0 +1,29 @@
...
+class ActionCable::Channel::BroadcastingTest < ActiveSupport::TestCase
+  class ChatChannel < ActionCable::Channel::Base
+  end
+
+  setup do
+    @connection = TestConnection.new
+  end
+
+  test "broadcasts_to" do
+    ActionCable.stubs(:server).returns mock().tap { |m| m.expects(:broadcast).with('action_cable:channel:broadcasting_test:chat:Room#1-Campfire', "Hello World")}
```
kaspth: Missing a space before the closing brace at the end of this line.

lib/action_cable/channel/naming.rb
kaspth: The doc is saying `controller_name` but should say `channel_name` :smile:

lib/action_cable/channel/streams.rb
kaspth: I originally thought the above sample could be expressed with `stream_for` but I see now that it can't. So yes, a separate example would be great :+1:

lib/action_cable/channel/broadcasting.rb
```diff
@@ -0,0 +1,29 @@
+module ActionCable
+  module Channel
+    module Broadcasting
+      extend ActiveSupport::Concern
+
+      included do
```
kaspth: We aren't wrapping the `delegate :channel_name, to: :class` in `naming.rb` in an included block, so why do we need it here?

kaspth:
@lsylvester starting to look good here :+1:

lib/action_cable/channel/streams.rb
lsylvester: @kaspth I have added example using `stream_for`/`broadcast_to` to the docs.

kaspth:
@lsylvester Well done :grin:

PR #35, opened 2015-07-22
Update gems and requires

twalpole:
Use latest faye-websocket,  celluloid, and websocket-driver (now that the issue with 0.6.0 is fixed).
Require 'rack/mockt' for Rack::MockRequest used in tests


kaspth:
Shouldn't this update the Gemfile lock too?

twalpole:
Whoops - too used to ignoring it for gem submissions

PR #36, opened 2015-07-22
Improve channel actions dispatcher to allow inheritance/mixins

### LINKED ISSUE #14: Improve channel action dispatcher to allow mixin and inheritance actions
Our security protection around client-side subscribers invoking actions on the server-side channel is too strict. It only allows for actions declared directly on the final channel class itself. Not methods included via mixins or ancestors in the inheritance tree. We solved the same problem in Action Controller, so let's port a version of that over here to allow all that is reasonable without opening up any attack vectors.

cristianbica:
Fixes #14

### DIFF
--- a/lib/action_cable/channel/base.rb
+++ b/lib/action_cable/channel/base.rb
@@ -75,6 +75,42 @@ module ActionCable
       attr_reader :params, :connection
       delegate :logger, to: :connection
 
+      class << self
+        # A list of method names that should be considered actions. This
+        # includes all public instance methods on a channel, less
+        # any internal methods (defined on Base), adding back in
+        # any methods that are internal, but still exist on the class
+        # itself.
+        #
+        # ==== Returns
+        # * <tt>Set</tt> - A set of all methods that should be considered actions.
+        def action_methods
+          @action_methods ||= begin
+            # All public instance methods of this class, including ancestors
+            methods = (public_instance_methods(true) -
+              # Except for public instance methods of Base and its ancestors
+              ActionCable::Channel::Base.public_instance_methods(true) +
+              # Be sure to include shadowed public instance methods of this class
+              public_instance_methods(false)).uniq.map(&:to_s)
+            methods.to_set
+          end
+        end
+
+        protected
+          # action_methods are cached and there is sometimes need to refresh
+          # them. ::clear_action_methods! allows you to do that, so next time
+          # you run action_methods, they will be recalculated
+          def clear_action_methods!
+            @action_methods = nil
+          end
+
+          # Refresh the cached action_methods when a new action_method is added.
+          def method_added(name)
+            super
+            clear_action_methods!
+          end
+      end
+
       def initialize(connection, identifier, params = {})
         @connection = connection
         @identifier = identifier
@@ -147,7 +183,7 @@ module ActionCable
         end
 
         def processable_action?(action)
-          self.class.instance_methods(false).include?(action)
+          self.class.action_methods.include?(action.to_s)
         end
 
         def dispatch_action(action, data)
@@ -167,7 +203,6 @@ module ActionCable
             end
           end
         end
-
 
         def run_subscribe_callbacks
           self.class.on_subscribe_callbacks.each { |callback| send(callback) }
--- a/test/channel/base_test.rb
+++ b/test/channel/base_test.rb
@@ -3,7 +3,22 @@ require 'stubs/test_connection'
 require 'stubs/room'
 
 class ActionCable::Channel::BaseTest < ActiveSupport::TestCase
-  class ChatChannel < ActionCable::Channel::Base
+  class ActionCable::Channel::Base
+    def kick
+      @last_action = [ :kick ]
+    end
+
+    def topic
+    end
+  end
+
+  class BasicChannel < ActionCable::Channel::Base
+    def chatters
+      @last_action = [ :chatters ]
+    end
+  end
+
+  class ChatChannel < BasicChannel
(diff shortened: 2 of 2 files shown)


cristianbica:
Mostly copy / pasted from rails' abstract_controller :)

kaspth:
Lets keep the newlines you deleted (to be consistent with the rest of the project) :smile:

test/channel/base_test.rb
```diff
@@ -92,6 +92,11 @@ def rm_rf
     assert_nil @channel.last_action
   end
 
+  test "try calling a public method defined on Base" do
+    @channel.perform_action 'action' => :unsubscribe_from_channel
+    assert_nil @channel.last_action
```
kaspth: Can you explain this test to me? I don't understand it.

cristianbica: Now that you opened my eyes it's pretty hard to do it :) Will add a real test for methods defined on `Base`

cristianbica:
@kaspth updated

lib/action_cable/channel/base.rb
```diff
@@ -75,6 +75,41 @@ class Base
...
+            methods = (public_instance_methods(true) -
+              # Except for public instance methods of Base and its ancestors
+              ActionCable::Channel::Base.public_instance_methods(true) +
+              # Be sure to include shadowed public instance methods of this class
+              public_instance_methods(false)).uniq.map(&:to_s)
+            methods.to_set
+          end
+        end
+
+        # action_methods are cached and there is sometimes need to refresh
```
dhh: What is this "sometimes" need? Just testing? If so, I don't think we need to expose this method in the public API.

cristianbica: Sometimes means that someone might access `action_methods` (and caching action methods)  before defining new methods. But I don't see any cases where one would call this manually (it's called automatically by `method_added`). I'm going to make this methods protected.

lib/action_cable/channel/base.rb
```diff
@@ -168,7 +204,6 @@ def action_signature(action, data)
           end
         end
 
-
```
kaspth: Keep this too :smile:

test/channel/base_test.rb
```diff
@@ -92,6 +111,23 @@ def rm_rf
     assert_nil @channel.last_action
   end
 
+  test "try calling a public method defined on Base" do
+    @channel.perform_action 'action' => :kick
+    assert_nil @channel.last_action
```
kaspth: So `kick` sets `@last_action`, but here we're asserting that last action is nil - is this supposed to mean that the action isn't allowed? I really don't understand it :smile:

lib/action_cable/channel/base.rb
cristianbica: why?

test/channel/base_test.rb
cristianbica: Yes, that's right. I used the same testing pattern used by @lifo [here](https://github.com/cristianbica/actioncable/blob/channel-action-dispatched/test/channel/base_test.rb#L109)

kaspth: I still think it's tough to grasp that the action is never reached. Maybe extending the test name a little would do the trick?

lib/action_cable/channel/base.rb
kaspth: Ah, sorry. Thought it was between a method scope change (like going from public to private), which is split by two newlines in Action Cable.

test/channel/base_test.rb
cristianbica: I've renamed the tests related to action methods to be more clear. 
I belive we still have and issue more related to the ways channels work as when the client sends an forbidden action the channel will log the request and do nothing. I think we might want to respond to the client with a forbidden action message.

kaspth: Way easier to understand those tests now, @cristianbica :+1:

Hm, I'm not sure. I can certainly see it would be nice in development if you happen to send a forbidden method. I don't see the use in production (those forbidden calls should already be long gone).

How would a forbidden response look like from the clients perspective?

cristianbica: Still don't have the big picture on actioncable and this discussion is out of scope of this PR. I will play more with actioncable in the coming days and come back with an issue or a PR if needed. Basically now it seems that whenever the client sends an action to the channel we assume that everything was ok on the server-side. Maybe it's a websockets conceptual thing or maybe developers should handle this logic. We will see :)

kaspth: Cool, looking forward to what you find :+1:

> Den 23/07/2015 kl. 14.27 skrev Cristian Bica <notifications@github.com>:
> 
> In test/channel/base_test.rb <https://github.com/rails/actioncable/pull/36#discussion_r35315566>:
> 
> > @@ -92,6 +111,23 @@ def rm_rf
> >      assert_nil @channel.last_action
> >    end
> >  
> > +  test "try calling a public method defined on Base" do
> > +    @channel.perform_action 'action' => :kick
> > +    assert_nil @channel.last_action
> Still don't have the big picture on actioncable and this discussion is out of scope of this PR. I will play more with actioncable in the coming days and come back with an issue or a PR if needed. Basically now it seems that whenever the client sends an action to the channel we assume that everything was ok on the server-side. Maybe it's a websockets conceptual thing or maybe developers should handle this logic. We will see :)
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/rails/actioncable/pull/36/files#r35315566>.
> 

--
Kasper

PR #38, opened 2015-07-24
Server::Base broadcasting test added

Axxiss:
Closing #30 as was being merged from my master branch instead a feature one. 

Rails and ApplicationCable are now stubbed.

I don't know what is the best way to handle config files required by the gem. As the configuration needs to load the following file `config/redis/cable.yml`.

At the moment I created an empty file under `test/config`

### DIFF
--- a/test/server/base_test.rb
+++ b/test/server/base_test.rb
@@ -1,0 +1,18 @@
+require 'test_helper'
+
+class ActionCable::Server::BaseTest < ActiveSupport::TestCase
+  setup do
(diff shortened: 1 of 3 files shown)


Axxiss:
Hi folks, any update on this?

dhh:
I can see how we ended up here, but this is basically almost just testing stubs the whole way through. I don't think that's that useful. I'd rather figure a way to have some system tests that can exercise against real redis etc.

Issue #40, opened 2015-07-26
Inconsistent version numbers


maclover7:
The [gemspec](https://github.com/rails/actioncable/blob/master/actioncable.gemspec#L3) lists that the current ActionCable version is 0.1.0. However, [action_cable.rb](https://github.com/rails/actioncable/blob/master/lib/action_cable.rb#L21) lists that the current ActionCable version is 0.0.3.

There should be one standardized place to record the version number, to avoid confusion.

matthewd:
Fixed by #41

PR #41, opened 2015-07-27
Move VERSION constant to version file and use this in the gemspec

craigsheen:
https://github.com/rails/actioncable/issues/40

### DIFF
--- a/lib/action_cable.rb
+++ b/lib/action_cable.rb
@@ -16,6 +16,7 @@ require 'em-hiredis'
 require 'redis'
 
 require 'action_cable/engine' if defined?(Rails)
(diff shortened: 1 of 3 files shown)


lib/action_cable.rb
```diff
@@ -18,8 +18,6 @@
 require 'action_cable/engine' if defined?(Rails)
 
 module ActionCable
-  VERSION = '0.0.3'
```
matthewd: We should require the new file, so the constant is still available to a running application.

craigsheen:
Updated to require the version file.

matthewd:
Thanks!

PR #42, opened 2015-07-28
Optimize channel class lookup

lsylvester:
Current implementation  calls `safe_constantize` at least once and possibly multiple times each time a subscription is added, which is a slow operation. 

Instead, this looks up the channel classes in a hash.

### DIFF
--- a/lib/action_cable/server/base.rb
+++ b/lib/action_cable/server/base.rb
@@ -36,11 +36,11 @@ module ActionCable
         @worker_pool ||= ActionCable::Server::Worker.pool(size: config.worker_pool_size)
       end
 
-      # Requires and returns an array of all the channel class constants in this application.
+      # Requires and returns an hash of all the channel class constants keyed by name.
       def channel_classes
         @channel_classes ||= begin
           config.channel_paths.each { |channel_path| require channel_path }
-          config.channel_class_names.collect { |name| name.constantize }
+          config.channel_class_names.each_with_object({}) { |name, hash| hash[name] = name.constantize }
         end
       end
 
--- a/test/connection/subscriptions_test.rb
+++ b/test/connection/subscriptions_test.rb
@@ -20,7 +20,7 @@ class ActionCable::Connection::SubscriptionsTest < ActiveSupport::TestCase
 
   setup do
     @server = TestServer.new
-    @server.stubs(:channel_classes).returns([ ChatChannel ])
+    @server.stubs(:channel_classes).returns(ChatChannel.name => ChatChannel)
 
     env = Rack::MockRequest.env_for "/test", 'HTTP_CONNECTION' => 'upgrade', 'HTTP_UPGRADE' => 'websocket'
     @connection = Connection.new(@server, env)
--- a/lib/action_cable/connection/subscriptions.rb
+++ b/lib/action_cable/connection/subscriptions.rb
@@ -24,9 +24,7 @@ module ActionCable
         id_key = data['identifier']
         id_options = ActiveSupport::JSON.decode(id_key).with_indifferent_access
 
-        subscription_klass = connection.server.channel_classes.detect do |channel_class|
-          channel_class == id_options[:channel].safe_constantize
-        end
+        subscription_klass = connection.server.channel_classes[id_options[:channel]]
 
         if subscription_klass
           subscriptions[id_key] ||= subscription_klass.new(connection, id_key, id_options)


kaspth:
We need a benchmark to verify your change has really improved performance :smile:

lsylvester:
@kaspth - here are the benchmarking scripts and results.

benchmarking scripts:
```ruby
# benchmark/apps/channels/one_channel.rb
class OneChannel < ActionCable::Channel::Base
end

# benchmark/apps/channels/two_channel.rb
class TwoChannel < ActionCable::Channel::Base
end

# benchmark/apps/channels/three_channel.rb
class ThreeChannel < ActionCable::Channel::Base
end

# benchmark/apps/channels/x_channel.rb
class XChannel < ActionCable::Channel::Base
end

# benchmark/add_benchmark.rb
require 'actioncable'
require 'benchmark'
require 'rack'
require 'rack/test'
class Connection < ActionCable::Connection::Base
  attr_reader :websocket
end

module Rails
  def self.logger
    @logger ||= ActiveSupport::TaggedLogging.new ActiveSupport::Logger.new(StringIO.new)
  end

  def self.root
    @root ||= Pathname.new(File.expand_path('..', __FILE__))
  end
end

module ApplicationCable
  class Connection < ActionCable::Connection::Base
  end
end


@server = ActionCable::Server::Base.new
env = Rack::MockRequest.env_for "/test", 'HTTP_CONNECTION' => 'upgrade', 'HTTP_UPGRADE' => 'websocket'
@connection = Connection.new(@server, env)

@subscriptions = ActionCable::Connection::Subscriptions.new(@connection)
@first_channel_identifier = { id: 1, channel: 'OneChannel' }.to_json
@last_channel_identifier= { id: 1, channel: 'XChannel' }.to_json

Benchmark.bm do |x|
  x.report("First"){ 10000.times{ @subscriptions.execute_command 'command' => 'subscribe', 'identifier' =>   @first_channel_identifier } }
  x.report("Last "){ 10000.times{ @subscriptions.execute_command 'command' => 'subscribe', 'identifier' =>   @last_channel_identifier } }
end

```
bundle exec ruby benchmark/add_benchmark.rb 
Before:
```
       user     system      total        real
First  0.110000   0.000000   0.110000 (  0.113642)
Last   0.160000   0.010000   0.170000 (  0.161453)
```

After:
```
       user     system      total        real
First  0.100000   0.010000   0.110000 (  0.102197)
Last   0.090000   0.000000   0.090000 (  0.096972)
```

lib/action_cable/server/base.rb
```diff
@@ -36,11 +36,11 @@ def worker_pool
         @worker_pool ||= ActionCable::Server::Worker.pool(size: config.worker_pool_size)
       end
 
-      # Requires and returns an array of all the channel class constants in this application.
-      def channel_classes
+      # Requires and returns an hash of all the channel class constants in this application.
+      def channel_classes_hash
```
dhh: Let's just keep the name `channel_classes`. It's fine to have that return a hash instead of an array.

dhh:
On top of the speed-up, I think this is just nicer code too.

test/connection/subscriptions_test.rb
```diff
@@ -20,7 +20,7 @@ def speak(data)
 
   setup do
     @server = TestServer.new
-    @server.stubs(:channel_classes).returns([ ChatChannel ])
+    @server.stubs(:channel_classes_hash).returns(ChatChannel.to_s => ChatChannel)
```
kaspth: Wouldn't `name` be clearer instead of `to_s`?

kaspth:
@lsylvester thanks for verifying it :+1:

lib/action_cable/server/base.rb
kaspth: Agreed.

I'd suggest some doc changes while we're here:

```ruby
# Requires and returns a hash of all channel class constants keyed by channel name.
```

(I'd leave out `in this application` as `channel_classes` returning classes for another application would be strange to me.)

lsylvester: I have updated the docs to say `channel class constants keyed by name`, I think that saying `keyed by channel name` could be confused with the `channel_name` method from #34

lsylvester:
@dhh @kaspth I have updated with your feedback.

kaspth:
Ah, you're right. Nice work :+1:

Issue #13, opened 2015-07-09
Protect against Cross-Site WebSocket Hijacking using the CRSF token


dhh:
WebSocket upgrade requests are susceptible to Cross-Site WebSocket Hijacking as detailed in https://www.christian-schneider.net/CrossSiteWebSocketHijacking.html. We need to use a similar CRSF token strategy as we do to protect the rest of a Rails application against a CRSF attack. Most likely, we can probably even use the same token, or a derivative thereof. But protect we must!

cristianbica:
I will give it a try

uu59:
CSRF token-like protection isn't needed I think.
Just verify request header's `Origin` field or common CORS protection using `Access-Control-Allow-Origin` by response header, they protect a user as well.
Browser always set `Origin` to request header when `new WebSocket("ws://example.com")` for negotiate the protocol.

Of course CSRF token-like mechanism works too, but using CORS headers approach is simpler than that.

gregmolnar:
@uu59 any HTTP header can be spoofed so I think a CSRF token would be a more secure solution.

jeremy:
You can't spoof the `Origin` header in browsers. Whitelisting `Origin` is a great (and much simpler) technique to prevent CSRF.

gregmolnar:
That's true, at least as longs as there is no SOP bypass vulnerability in the browser(which is rare).

cristianbica:
I'm in progress with the CSRF implementation. An `Origin` based implementation would be much simpler.  I'm going to explore that and we'll see how it looks.

gregmolnar:
I think simple is better and SOP should be the browsers responsibility. Thanks for your work on this @cristianbica !

jeremy:
Origin whitelist implemented at #85, moving there!

PR #43, opened 2015-07-28
CSWSH protection

cristianbica:
This is my first stab at implementing connection forgery protection for websockets.
some notes:
 - the session part isn't what I had in mind but I haven't found another solution
 - some code is copied from controllers forgery protection; when merged into rails we dry-up the code a little

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -102,6 +102,3 @@ DEPENDENCIES
   mocha
   puma
   rake
-
-BUNDLED WITH
-   1.10.5
--- a/README.md
+++ b/README.md
@@ -245,6 +245,36 @@ For a full list of all configuration options, see the `ActionCable::Server::Conf
 
 Also note that your server must provide at least the same number of database connections as you have workers. The default worker pool is set to 100, so that means you have to make at least that available. You can change that in `config/database.yml` through the `pool` attribute.
 
+## Security
+
+Websockets default implementation is vulnerable to [Cross-Site WebSocket Hijacking](https://www.christian-schneider.net/CrossSiteWebSocketHijacking.html)
+attacks. Action Cable has a protection mechanism that uses the CSRF token from Controllers.
+
+To activate the protection you'll need to add `protect_from_hijacking` to your `ActionCable::Connection::Base` subclass:
+
+```ruby
+module ApplicationCable
+  class Connection < ActionCable::Connection::Base
+    protect_from_hijacking
+  end
+end
+```
+
+and add the `csrf_meta_tags` to your layout
+```html
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Action Cable Examples</title>
+  <%= csrf_meta_tags %>
+  <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': true %>
+  <%= javascript_include_tag 'application', 'data-turbolinks-track': true %>
+</head>
+```
+
+Important: Action Cable needs to have access to the meta tags so make sure that the `csrf_meta_tags` is above the `javascript_link_tag` or the javascript / coffescript creating the cable is executed on a domready event.
+
+
 ## Starting the cable server
 
 As mentioned, the cable server(s) is separated from your normal application server. It's still a rack application, but it is its own rack
--- a/lib/action_cable/connection/base.rb
+++ b/lib/action_cable/connection/base.rb
@@ -12,16 +12,16 @@ module ActionCable
     #   module ApplicationCable
     #     class Connection < ActionCable::Connection::Base
     #       identified_by :current_user
-    #   
-    #       def connect
+    #
+    #       after_connect do
     #         self.current_user = find_verified_user
     #         logger.add_tags current_user.name
     #       end
     #
-    #       def disconnect
+    #       after_disconnect do
     #         # Any cleanup work needed when the cable connection is cut.
     #       end
-    #   
+    #
     #       protected
     #         def find_verified_user
     #           if current_user = User.find_by_identity cookies.signed[:identity_id]
@@ -47,6 +47,8 @@ module ActionCable
       include Identification
       include InternalChannel
       include Authorization
+      include Callbacks
+      include HijackingProtection
 
       attr_reader :server, :env
       delegate :worker_pool, :pubsub, to: :server
@@ -75,14 +77,14 @@ module ActionCable
           websocket.on(:open)    { |event| send_async :on_open   }
           websocket.on(:message) { |event| on_message event.data }
           websocket.on(:close)   { |event| send_async :on_close  }
-          
+
           respond_to_successful_request
         else
           respond_to_invalid_request
         end
       end
 
-      # Data received over the cable is handled by this method. It's expected that everything inbound is encoded with JSON. 
+      # Data received over the cable is handled by this method. It's expected that everything inbound is encoded with JSON.
       # The data is routed to the proper channel that the connection has subscribed to.
       def receive(data_in_json)
         if websocket.alive?
@@ -130,6 +132,17 @@ module ActionCable
           request.cookie_jar
         end
 
+        # The session of the request that initiated the websocket connection. Useful for performing server-side validations.
+        def session
+          @session ||= begin
+            if defined?(Rails.application) && Rails.application
+              Rails.application.config.session_store.new(request, Rails.application.config.session_options).load_session(request.env).last
+            else
+              request.session
+            end
+          end
+        end
+
 
       private
         attr_reader :websocket
@@ -138,7 +151,8 @@ module ActionCable
         def on_open
           server.add_connection(self)
 
-          connect if respond_to?(:connect)
+          run_callbacks :connect
+
           subscribe_to_internal_channel
           heartbeat.start
 
@@ -161,7 +175,7 @@ module ActionCable
           unsubscribe_from_internal_channel
           heartbeat.stop
 
-          disconnect if respond_to?(:disconnect)
+          run_callbacks :disconnect
         end
 
 
@@ -177,7 +191,7 @@ module ActionCable
 
         # Tags are declared in the server but computed in the connection. This allows us per-connection tailored tags.
         def new_tagged_logger
-          TaggedLoggerProxy.new server.logger, 
+          TaggedLoggerProxy.new server.logger,
             tags: server.config.log_tags.map { |tag| tag.respond_to?(:call) ? tag.call(request) : tag.to_s.camelize }
         end
 
--- a/lib/action_cable/connection/callbacks.rb
+++ b/lib/action_cable/connection/callbacks.rb
@@ -1,0 +1,61 @@
+module ActionCable
+  module Connection
+    # = Action Cable Callbacks
+    #
+    # Action Cable provides hooks during the life cycle of a connection.
+    # Callbacks allow you to trigger logic during the life cycle of a
+    # connection. Available callbacks are:
+    #
+    # * <tt>after_connect</tt>
+    # * <tt>after_disconnect</tt>
+    #
+    module Callbacks
+      extend  ActiveSupport::Concern
+      include ActiveSupport::Callbacks
+
+      included do
+        define_callbacks :connect, scope: :name
+        define_callbacks :disconnect, scope: :name
+      end
+
+      module ClassMethods
+        # Defines a callback that will get called right after the
+        # connection is established.
+        #
+        #   module ApplicationCable
+        #     class Connection < ActionCable::Connection::Base
+        #       identified_by :current_user
+        #       after_connect :prepare_connection
+        #
+        #       def prepare_connection
+        #         self.current_user = find_verified_user
+        #         logger.add_tags current_user.name
+        #       end
+        #     end
+        #   end
+        #
+        def after_connect(*filters, &blk)
+          set_callback(:connect, :before, *filters, &blk)
+        end
+
+        # Defines a callback that will get called right after the
+        # connection is closed.
+        #
+        #   module ApplicationCable
+        #     class Connection < ActionCable::Connection::Base
+        #       identified_by :current_user
+        #       after_disconnect :cleanup
+        #
+        #       def cleanup
+        #         # Any cleanup work needed when the cable connection is cut.
+        #       end
+        #     end
+        #   end
+        #
+        def after_disconnect(*filters, &blk)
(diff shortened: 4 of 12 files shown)


lib/assets/javascripts/cable.js.coffee
```diff
@@ -5,4 +5,8 @@
   PING_IDENTIFIER: "_ping"
 
   createConsumer: (url) ->
+    token = $("meta[name=csrf-token]").attr("content")
```
lsylvester: The current documentation for creating a consumer has it created immediately (ie, not when DOM is ready or similar) and by default in a newly generated rails app the tokens appear after the javascript include tag, so by default the $("meta[name=csrf-token]") will not exist when this code executes.

I think that we will need to either:
* update docs to create the consumer after the DOM is ready, or
* update docs to include moving `csrf_meta_tags` to before `javascript_include_tag`

I think the second option is better because it would mean that the consumer would be set up earlier.

Also - This adds a dependency on jquery (or other javascript library with `$`). Not sure if we should try to avoid the dependency but we should at least document it if we have it.

javan: :+1: to both points. Can do `document.querySelector("meta[name=csrf-token]")?.getAttribute("content")` instead.

lib/assets/javascripts/cable.js.coffee
```diff
@@ -5,4 +5,8 @@
...
+    if token
+      separator = if url.indexOf("?")>=0 then "&" else "?"
```
javan: Need some spaces around the `>=`. And it might be nice to wrap up all this URL wrangling in a helper function.

lib/assets/javascripts/cable.js.coffee
cristianbica: @lsylvester  Yep that is a valid case. In my test with http://github.com/rails/actioncable-examples I had to move  the `csrf_meta_tags` before `javascript_include_tag`. I think we should move the cable connection creating code in a domready event. We can do that without jquery as browsers which support websockets support also DOMContentLoaded (see [this](http://caniuse.com/#feat=websockets), [this](http://caniuse.com/#feat=domcontentloaded) and [this](http://beeker.io/jquery-document-ready-equivalent-vanilla-javascript)).
@javan will do that

lib/assets/javascripts/cable.js.coffee
cristianbica: will do. thanks

cristianbica:
@lsylvester @javan updated

cristianbica:
@dhh @kaspth can you review this? thanks

Gemfile.lock
```diff
@@ -102,6 +102,3 @@ DEPENDENCIES
   mocha
   puma
   rake
-
-BUNDLED WITH
-   1.10.5
```
kaspth: We should leave this in.

README.md
```diff
@@ -245,6 +245,36 @@ For a full list of all configuration options, see the `ActionCable::Server::Conf
 
 Also note that your server must provide at least the same number of database connections as you have workers. The default worker pool is set to 100, so that means you have to make at least that available. You can change that in `config/database.yml` through the `pool` attribute.
 
+## Security
+
+Websockets default implementation is vulnerable to [Cross-Site WebSocket Hijacking](https://www.christian-schneider.net/CrossSiteWebSocketHijacking.html)
```
kaspth: WebSocket is camel cased.

kaspth: Can we write a bit about what the hijacking is, why we need protection, and maybe why it's safe to reuse the controller tokens?

README.md
```diff
@@ -245,6 +245,36 @@ For a full list of all configuration options, see the `ActionCable::Server::Conf
...
+attacks. Action Cable has a protection mechanism that uses the CSRF token from Controllers.
```
kaspth: I'd change `Controllers` to `Action Controller controllers`

README.md
```diff
@@ -245,6 +245,36 @@ For a full list of all configuration options, see the `ActionCable::Server::Conf
...
+
+```ruby
+module ApplicationCable
+  class Connection < ActionCable::Connection::Base
+    protect_from_hijacking
+  end
+end
+```
+
+and add the `csrf_meta_tags` to your layout
```
kaspth: And with capital a, add a colon and a newline.

README.md
```diff
@@ -245,6 +245,36 @@ For a full list of all configuration options, see the `ActionCable::Server::Conf
...
+<html>
+<head>
+  <title>Action Cable Examples</title>
+  <%= csrf_meta_tags %>
+  <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': true %>
+  <%= javascript_include_tag 'application', 'data-turbolinks-track': true %>
+</head>
+```
+
+Important: Action Cable needs to have access to the meta tags so make sure that the `csrf_meta_tags` is above the `javascript_link_tag` or the javascript / coffescript creating the cable is executed on a domready event.
```
kaspth: The README explains the default client side cable setup as:

```coffeescript
# app/assets/javascripts/application_cable.coffee
#= require cable

@App = {}
App.cable = Cable.createConsumer "ws://cable.example.com"
```

So I don't think we should mention the dom ready execution.

I'd just reprase as:

```markdown
Note: For Action Cable to have access to the meta tags, they should be inserted before the `javascript_link_tag`.
```

lib/action_cable/connection/base.rb
```diff
@@ -12,16 +12,16 @@ module Connection
     #   module ApplicationCable
     #     class Connection < ActionCable::Connection::Base
     #       identified_by :current_user
-    #   
-    #       def connect
+    #
+    #       after_connect do
```
kaspth: I'm not sure I agree with turning `connect` and `disconnect` into a callback. As methods they're nice anchors which give feeling of how connection (or disconnection) happens.

If it's all callbacks it loses meaning. I.e. imagine if Active Record's actual saving was just implemented as the last `before_save` or the first `after_save`, sounds like that would be complex down the road.

Though I'm definitely :+1: on adding before and after callbacks, I think we should keep the `connect` and `disconnect` methods.

kaspth: Ah, I hadn't seen that there were only `after` callbacks. Then it makes even less sense to get rid of the methods for me.

lib/action_cable/connection/callbacks.rb
```diff
@@ -0,0 +1,61 @@
+module ActionCable
+  module Connection
+    # = Action Cable Callbacks
+    #
+    # Action Cable provides hooks during the life cycle of a connection.
+    # Callbacks allow you to trigger logic during the life cycle of a
```
kaspth: Lets rewrite this sentence and the one before it to one sentence - they mostly repeat each other.

lib/action_cable/connection/callbacks.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+
+      module ClassMethods
+        # Defines a callback that will get called right after the
+        # connection is established.
+        #
+        #   module ApplicationCable
+        #     class Connection < ActionCable::Connection::Base
+        #       identified_by :current_user
+        #       after_connect :prepare_connection
```
kaspth: So after we've connected, we prepare the connection? It sounds like `prepare_connection` doesn't make sense as an `after_connect` callback.

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
+require 'active_support/security_utils'
+
+module ActionCable
+  module Connection
+    # Action Cable connection are protected from Cross-Site WebSocket Hijacking
```
kaspth: `The Action Cable connection is`?

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+    # (CSWSH) by using the same CSRF token protecting Rails' controllers.
+    # When starting a Websocket connection the Action Cable javascript will
```
kaspth: `WebSocket` and `JavaScript` :smile:

kaspth: Actually, I'd rephrase as:

```ruby
# When starting a connection Action Cable sends along the CSRF token. The token is...
```

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+    # send along the CSRF token used to protect controllers. The token is
+    # stored as a random string in the session, to which an attacker does not
+    # have access. When a Websocket connection request reaches your
+    # application, Action Cable verifies the received token with the token in
+    # the session.
+    #
+    # CSWSH protection is turned on with the <tt>protect_from_hijacking</tt>
+    # method.
+    #
+    # module ApplicationCable
```
kaspth: Should be indented by two more spaces.

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+    #   class Connection < ActionCable::Connection::Base
+    #     protect_from_hijacking, if: -> { current_user.guest? }
+    #   end
+    # end
+    #
+    # Make sure that every page that requires starting a Websocket connection
```
kaspth:
```ruby
# Any page starting a connection must include the CSRF token by outputting
# <tt>csrf_meta_tags</tt> in the HTML +head+.
```

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+    # includes in it's head the CSRF. This is accomplised by including
+    # <tt>csrf_meta_tags</tt> in the HTML +head+. Action Cable needs to have
+    # access to the meta tags so make sure that the <tt>csrf_meta_tags</tt> is
+    # above the <tt>javascript_link_tag</tt> or the javascript / coffescript
+    # creating the cable is executed on a domready event.
```
kaspth: Same thing as mentioned earlier about `domready`.

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+    # {Cross-Site WebSocket Hijacking}[https://www.christian-schneider.net/CrossSiteWebSocketHijacking.html].
+    # {Ruby on Rails Security Guide}[http://guides.rubyonrails.org/security.html].
+    module HijackingProtection
+      extend ActiveSupport::Concern
+
+      AUTHENTICITY_TOKEN_LENGTH = 32
+
+      module ClassMethods
+        def protect_from_hijacking(options = {})
+          options = options.reverse_merge(prepend: true)
```
kaspth: Why not just inline the `reverse_merge` and skip resetting the variable?

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+          after_connect :validate_connection, options
+        end
+      end
+
+
+      protected
+      # This is the actual method verifying the token
```
kaspth: Why not just rename it to someting like `verify_connection_uses_valid_token` and skip the comment?

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+      def validate_connection
+        reject_unauthorized_connection "Invalid CSWSH/CSRF token" unless valid_authentication_token?(request.params["token"])
+      end
+
+      def valid_authentication_token?(encoded_masked_token)
+        if encoded_masked_token.nil? || encoded_masked_token.empty? || !encoded_masked_token.is_a?(String)
```
kaspth: Can't we use `blank?` instead of `nil? || `empty?`?

We also don't care about returning singletons, so we could simplify as:

```ruby
return unless encoded_masked_token.present? && encoded_masked_token.is_a?(String)
```

lib/action_cable/connection/hijacking_protection.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        begin
+          masked_token = Base64.strict_decode64(encoded_masked_token)
+        rescue ArgumentError # encoded_masked_token is invalid Base64
+          return false
+        end
+
+        if masked_token.length == AUTHENTICITY_TOKEN_LENGTH
+          # This is actually an unmasked token. This is expected if
+          # you have just upgraded to masked tokens, but should stop
```
kaspth: Do we still need this check? Or can we assume that people on Rails 5, will have hit this path long ago?

lib/assets/javascripts/cable.js.coffee
```diff
@@ -5,4 +5,11 @@
   PING_IDENTIFIER: "_ping"
 
   createConsumer: (url) ->
-    new Cable.Consumer url
+    new Cable.Consumer @attachCsrfToken(url)
+
+  attachCsrfToken: (url) ->
```
kaspth: `attachCSRFToken`?

lib/assets/javascripts/cable.js.coffee
```diff
@@ -5,4 +5,11 @@
...
+    token = document.querySelector("meta[name=csrf-token]")?.getAttribute("content")
+    if token
+      separator = if url.indexOf("?")>=0 then "&" else "?"
```
kaspth: What about putting a space before and after `>=`?

kaspth:
@cristianbica be careful what you wish for :sweat_smile: Comments notwithstanding, great work, :+1:.

We also need a rebase.

cristianbica:
Going through the emails I was thinking at the same thing :). Thanks. Will
work on the feedback in the coming days

-- 
Cristian Bica

cristianbica:
Hey sorry guys but I've been pretty busy lately.
I'm going to put this PR on hold for now until we decide on a solution (discussion here https://github.com/rails/actioncable/issues/13#issuecomment-145926704)

jeremy:
Origin whitelisting notwithstanding, supporting other sorts of pre-connect checks would be great. We should be able to plug in token-based CSRF protection without much pain. Worth moving this forward to support that.

jeremy:
(Re. #85)

jeremy:
@cristianbica Let's reopen this against rails mainline when/if you choose to pursue it further.

cristianbica:
@jeremy been a bit busy lately but if no one picks this I'll give it a spin early January. Thanks

PR #50, opened 2015-07-30
update docs for broadcast_to to pass a message.

lsylvester:


lib/action_cable/channel/streams.rb
```diff
@@ -35,7 +35,7 @@ module Channel
     #
     # You can then broadcast to this channel using:
     #
-    #   CommentsChannel.broadcast_to(@post)
+    #   CommentsChannel.broadcast_to(@post, @comment.as_json)
```
kaspth: `broadcast` already calls `to_json` (https://github.com/rails/actioncable/blob/ae61460cb2ebcc25528686d896c136522690f083/lib/action_cable/server/broadcasting.rb#L47), so we just need `@comment` here.

lsylvester: @kaspth Updated.

Issue #47, opened 2015-07-30
ActionCable::Channel::Base to subclass AbstractController::Base (RFC)


lsylvester:
The action processing of `ActionCable::Channel::Base` seems to be similar to the generic action processing in `AbstractController`.

Is there any reason why `ActionCable::Channel::Base` couldn't/shouldn't be a subclass of `AbstractController::Base` to get its action processing implementation?

lsylvester:
Implemented in #52 - any discussion can happen there.

PR #52, opened 2015-07-31
Use Abstract Controller for Channels

lsylvester:
Use `AbstractController::Base` to get the action dispatching for `ActionChannel::Base` as described in #47

### DIFF
--- a/lib/action_cable/channel/base.rb
+++ b/lib/action_cable/channel/base.rb
@@ -1,3 +1,5 @@
+require 'abstract_controller'
+
 module ActionCable
   module Channel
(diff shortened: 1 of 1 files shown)


kaspth:
Already there are lines drawn between channels and controllers elsewhere - we say they are alike in some way. Though I'm not sure if we should go all the way and subclass from `AbstractController::Base`, that might set the wrong precedent down the line.

For instance, what if channels evolve to be less like controllers? Admittedly that's a way off, but I think it's tougher to change a superclass versus just extracting the action method logic to a concern and including that in both places.

Either way, I'd rather postpone this decision altogether till Action Cable is merged in to Rails as we have something that works now.

Thanks for all the great work on Action Cable, @lsylvester :heart:

PR #57, opened 2015-08-07
Add logging to stdout in development mode.

lsylvester:


ascrazy:
In my opinion there should be ability to use stdout logging in production mode too, in order to match heroku's [12 factor app manifest](http://12factor.net/logs).

> A twelve-factor app never concerns itself with routing or storage of its output stream. It should not attempt to write to or manage logfiles. Instead, each running process writes its event stream, unbuffered, to stdout

lsylvester:
@ascrazy I think that to log to stdout in production you would configure the `Rails.logger` to log to stdout instead of a file. This PR is adding logging to stdout in addition to the standard log file.

PR #58, opened 2015-08-10
Clear out the streams when they are stopped.

lsylvester:
When `stop_all_streams` is called, the streams are left in the channels `streams`. So then next time `stop_all_streams` is called, it will try and stop the same streams again. 

It looks like that if a channel is designed to persist the entire duration of a users session, it will continue to accumulate more and more streams.

This PR adds clears out the streams after stoping them, to prevent a accumulation of inactive streams.

### DIFF
--- a/lib/action_cable/channel/streams.rb
+++ b/lib/action_cable/channel/streams.rb
@@ -88,7 +88,7 @@ module ActionCable
         streams.each do |broadcasting, callback|
           pubsub.unsubscribe_proc broadcasting, callback
           logger.info "#{self.class.name} stopped streaming from #{broadcasting}"
-        end
+        end.clear
       end
 
       private


lib/action_cable/channel/streams.rb
```diff
@@ -89,6 +89,7 @@ def stop_all_streams
           pubsub.unsubscribe_proc broadcasting, callback
           logger.info "#{self.class.name} stopped streaming from #{broadcasting}"
         end
```
dhh: You can just do `end.clear` instead. `collect` returns itself afterwards.

lsylvester: @dhh Thanks for your feedback. This has been updated.

PR #62, opened 2015-08-18
Add contributing file

maclover7:
cc @dhh


dhh:
Nice. Is it actually true that we already have "hundreds of contributors"?

maclover7:
Lol - am just using the same template across all Rails sub projects. There are only 19 contributors to ActionCable, including yourself and myself.

dhh:
Then maybe just go with "the many" or "lots" or something like that.

matthewd:
I don't think we need to do this; actioncable will be merged into rails/rails soon enough, and the other components don't have separate contributing files.

dhh:
Even better.

Issue #63, opened 2015-08-19
Identifiers that don't implement to_global_id


mrlhumphreys:
Hi everyone,

I was working on replacing polling with `ActionCable` in an app. It's been good so far and things are working mostly as expected. Thanks everyone who worked on this. I've been anticipating this kind of feature for a long time.

I ran into a small issue when using a uuid string as the identifier for a connection. (The scenario is that this uuid string is a unique identifier for a session and the user is not logged in, so current_user is nil)

An error was raised in lib/connection/identification.rb#connection_gid saying:

```
undefined method `to_global_id' for "121cce49-3f9a-4871-904f-5b6d0e0f5d25":String
```

So, it seems like there's a requirement that the identifier should be something that implements `to_global_id`. Currently, I'm working around it by using a hand-rolled class that implements `to_global_id`.

My question is, should the identifier be able to handle objects that don't implement `to_global_id`?

If so, then this leads to the next question on how a particular object can be converted to a string to be used as the `connection_identifier`. `to_global_id` is fine for `ActiveRecord` models. But, for other kinds of objects, I'm not sure.

Any thoughts?

dhh:
Yeah, we should probably just try(:to_global_id) and then do try(:to_s) if that doesn't work 👍

mrlhumphreys:
Thanks. That sounds good. I'll work on implementing that.

kaspth:
Fixed by https://github.com/rails/actioncable/pull/65

PR #65, opened 2015-08-21
Support connection identifiers that don't implement to_global_id

mrlhumphreys:
This is related to #63 

Previously, issues would occur when constructing the `connection_gid` in `ActionCable::Connection::Identification` as it would attempt to call `to_global_id` on the identifier even if it didn't have such a method. 

This change allows us to use identifiers that don't implement `to_global_id`.

Any suggestions welcomed.

### DIFF
--- a/lib/action_cable/connection/identification.rb
+++ b/lib/action_cable/connection/identification.rb
@@ -27,7 +27,7 @@ module ActionCable
 
       private
         def connection_gid(ids)
-          ids.map { |o| o.to_global_id.to_s }.sort.join(":")
+          ids.map { |o| (o.try(:to_global_id) || o).to_s }.sort.join(":")
         end
     end
   end
--- a/test/connection/string_identifier_test.rb
+++ b/test/connection/string_identifier_test.rb
@@ -1,0 +1,39 @@
+require 'test_helper'
+require 'stubs/test_server'
+
+class ActionCable::Connection::StringIdentifierTest < ActiveSupport::TestCase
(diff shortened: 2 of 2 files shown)


lib/action_cable/connection/identification.rb
```diff
@@ -27,7 +27,7 @@ def connection_identifier
 
       private
         def connection_gid(ids)
-          ids.map { |o| o.to_global_id.to_s }.sort.join(":")
+          ids.map { |o| (o.try(:to_global_id) || o.try(:to_s)).to_s }.sort.join(":")
```
kaspth: Why do we need to wrap `to_s` in `try`? If it's to ferret out nil values that's undone by the second `to_s`.

This should probably just be:

```ruby
ids.map { |o| (o.try(:to_global_id) || o).to_s }.sort.join(":")
```

kaspth: What I mean by undone is that `nil.try(:to_s)` returns `nil` and then `nil.to_s` returns an empty string.

kaspth:
Minor quibble, but :+1: overall

lib/action_cable/connection/identification.rb
mrlhumphreys: That makes sense. I have removed the superfluous to_s.

kaspth:
Thanks :grin:

As a future tip, if you add "Fixes https://github.com/rails/actioncable/pull/65" to your PR description it'll automatically close the issue when the PR is merged.

PR #68, opened 2015-08-23
added support for kqueue

sespindola:
Hi, 
 I added a little ternary switch in order to use Kqueue on BSD platforms.


sespindola:
An even better solution was committed in 98855fea634aa1c427c18841cf14fa3777201c4b.
Closing.

Issue #32, opened 2015-07-20
Contexts/Callbacks/Hook ideas (RFC)


mieko:
I'm porting an application from Rails 4 + a homegrown Faye mess to Rails 5-pre + ActionCable.  I was probably the most excited person on earth when this repo went public.

This RFC is about Connection/Channel lifecycle hooks generally, but I'll illustrate with my actual use-case.

The application is heavily tenanted, with a model that works a lot like the Apartment gem.  Outside of a few well-defined areas, all of our logic looks like normal Rails practices.  Models, Controllers, Jobs, etc, are written as if nothing weird with tenancy is going on behind the scenes.  This is possible because there tends to be a publicly-documented bottleneck I can use to swap the ground beneath the code for each of the frameworks, things like this: https://gist.github.com/mieko/13a312b10b6087f38a87

I'm trying to find a clean way to extend this behavior to ActionCable.  To cover all cases, I've had to end up with something like the following (paraphrased for clarity):

```ruby
module ApplicationCable
  class Connection < ActionCable::Connection::Base
    include ActivateTenant
    identified_by :current_tenant, :current_user

    def connect
      # Pull the tenant out of the hostname
      self.current_tenant = tenant_for_request(request)
      self.current_user = user_for_request(request)
    end
  end
end
```

And here's where it gets hacky:

```ruby
module ApplicationCable
  class Channel < ActionCable::Channel::Base
    # For sake of exposition, each method written out 
    protected
    def dispatch_action(*); current_tenant.activate! { super }; end
    def run_subscribe_callbacks(*); current_tenant.activate! { super }; end
    def run_unsubscribe_callbacks(*); current_tenant.activate! { super }; end
  end
end
```

This seems(?) to be required because ActionCable doesn't have the "full-featured" callback handling of something like ActionController.  `on_{un,}subscribe` would be perfect for this, but AC::C::Base registers `subscribed` itself for instances, and registering further callbacks append to the chain.  This means `ApplicationCable::Channel` cannot register an `on_subscribe` that'd execute before subclass instances' `subscribed`.

Without going down the rabbit hole of `prepend_...` and `append_...`, I think it'd be nice if something like the following were to work out of the box:

```ruby
module ApplicationCable
  class Channel < ActionCable::Channel::Base
    identified_by ...
    on_subscribe :activate_tenant
    
    protected
    def activate_tenant
      current_tenant.activate!(...)
    end
  end
end

class AppearanceChannel < ApplicationCable::Channel
[...]

PR #69, opened 2015-08-24
Use ActiveSupport::Callbacks for Channel subscription callbacks.

mieko:
(As mentioned in #32)

 * Rely on AS::Callbacks for callback handling.
 * Add before_subscribe, after_subscribe, before_unsubscribe and
   after_unsubscribe convenience methods
 * alias on_subscribe and on_unsubscribe to after_subscribe and
   after_unsubscribe, respectively.
 * Remove `subscribed` and `unsubscribed` from the callback chain:
   these methods are now executed as the *subject* of the callbacks.
 * Update portions of ActionCable to use the more specific
   callback names.

### DIFF
--- a/lib/action_cable/channel/base.rb
+++ b/lib/action_cable/channel/base.rb
@@ -71,9 +71,6 @@ module ActionCable
       include Naming
       include Broadcasting
 
-      on_subscribe   :subscribed
-      on_unsubscribe :unsubscribed
-
       attr_reader :params, :connection
       delegate :logger, to: :connection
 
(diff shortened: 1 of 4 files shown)


antifun:
:+1: for this; we are in a similar situation, and would love to see this get in. How can we help?

mieko:
The PR would probably need tests and documentation, but I found myself short of time.  I'll try to push updates to the branch as my schedule allows over the next week or so.

dhh:
Seems nice not to reimplement callbacks :+1: 

@lifo thoughts?

lifo:
Makes sense to me.

PR #70, opened 2015-08-30
Сlient refactoring in order to support CommonJS modules

ascrazy:
Since more and more people choose CommonJS or ES6 modules for resolving dependencies in client-side codebase, I think that actioncable client should support CommonJS modules, and should provide npm package for fast and easy integration with modern asset bundlers, like browserify or webpack.

I've done a little work on support for CommonJS modules (compatibility with sprockets has saved).
I've also added a few tests to ensure that these changes haven't broke top-level module API's.

With this changes client-side code doesn't look pretty clear, so I think we should review ability to move development of client to CommonJS only modules, and include to gem already pre-compiled version which will be compatible with sprockets.

[Vagrantfile](https://gist.github.com/ascrazy/12395e5daa3e55360deb)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,13 @@
 test/tests.log
+
+node_modules/
+npm-debug.log
+
+dist/*
+!dist/.keep
+
+test/sprockets/cable.js
+
+.vagrant/
+bootstrap.sh
+Vagrantfile
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,10 @@
+lib/action_cable
+lib/*.rb
+test/
+actioncable.gemspec
+Rakefile
+Gemfile
+Gemfile.lock
+.ruby-version
+bootstrap.sh
+Vagrantfile
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,6 +84,8 @@ GEM
       thor (>= 0.18.1, < 2.0)
     rake (10.4.2)
     redis (3.2.1)
+    sprockets (3.3.3)
+      rack (~> 1.0)
     thor (0.19.1)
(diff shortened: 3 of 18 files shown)


ascrazy:
I am working on CommonJS-only solution in separate branch.
There client code lives in separate folder with tests and other stuff, and is copied to gem sources in build-hook.

[See diff with details](https://github.com/rails/actioncable/compare/rails:master...ascrazy:commonjs-modules)

javan:
Hi @ascrazy, I appreciate the work you've done here and I'm sympathetic to providing modules for other build systems. I don't think a change like this belongs here though. Action Cable's main purpose is to provide an integrated WebSocket system for Rails, and as such it uses Rails' integrated asset manager (Sprockets). If we are going support JavaScript modules, I think it should be handled by Sprockets so that all other libraries (like Turbolinks) can support them too. Closing for now.

ascrazy:
Hi @javan, 
But this solution retains sprockets support, why we can't support two build systems from one codebase? I am very interested in using actioncable in projects where rails is used as backend, and frontend development lives  entirely in npm ecosystem, but now there are no ways to do it, without copying client files to frontend project, and of course this is very sad.

javan:
I'm fine with providing a pre-built .js file for consumption, but I'm not hot on side-stepping Sprockets as the main asset manager. Changing `#= require ...` to `require(...)` and conditionally setting the global in each file are a significant change from the normal Sprockets workflow.

Would you be able to work with a pre-built action_cable.js file with `Cable` set as a global? Alternatively, you could have your Rails app serve Action Cable's JavaScript and load it from your front-end in its own `<script>` tag.

PR #73, opened 2015-09-04
Easy websocket url configuration

### LINKED ISSUE #72: Management of client (JS-side) configuration for WS server
So in your JS "client" you have to call something like `Cable.createConsumer('ws://example.com:12345')` to create your App.cable. But in development you probably want to call something like `Cable.createConsumer('ws://127.0.0.1:12345')` instead. Not sure how it'd work, but maybe there needs to be some configuration management for the JS side of things? 

The only idea I have is a .yml or .json file somewhere that causes the js asset pipeline to create/include some sort of `App.settings` or `Rails.settings` object of config key/vals based on the current environment so that you can just call something like `Cable.createConsumer(App.settings.websocket_host)` and have it be the right host for the env.

Anyway, just some thoughts. There are probably other approaches that make sense.

smellsblue:
This fixes #72 by introducing a url configuration that is used by a meta tag similar to the CSRF meta tags, allowing Cable.createConsumer to get the correct websicket url from your environment config.

This is basically how I am solving this problem in my app, but I'd be happy to work on it more if a different solution is desirable.

Also would be happy to do any cleanup that might need done.

### DIFF
--- a/lib/assets/javascripts/cable.coffee.erb
+++ b/lib/assets/javascripts/cable.coffee.erb
@@ -4,5 +4,9 @@
 @Cable =
   INTERNAL: <%= ActionCable::INTERNAL.to_json %>
 
-  createConsumer: (url) ->
+  createConsumer: (url = @getConfig("url")) ->
     new Cable.Consumer url
+
+  getConfig: (name) ->
+    element = document.head.querySelector("meta[name='action-cable-#{name}']")
+    element?.getAttribute("content")
--- a/README.md
+++ b/README.md
@@ -330,6 +330,26 @@ ActionCable.server.config.log_tags = [
 ]
 ```
 
+Your websocket url might change between environments. If you host your production server via https, you will need to use the wss scheme
+for your ActionCable server, but development might remain http and use the ws scheme. You might use localhost in development and your
+domain in production. In any case, to vary the websocket url between environments, add the following configuration to each environment:
+
+```ruby
+config.action_cable.url = "ws://example.com:28080"
+```
+
+Then add the following line to your layout before your JavaScript tag:
+
+```erb
+<%= action_cable_meta_tag %>
+```
+
+And finally, create your consumer like so:
+
+```coffeescript
+App.cable = Cable.createConsumer()
+```
+
 For a full list of all configuration options, see the `ActionCable::Server::Configuration` class.
 
 Also note that your server must provide at least the same number of database connections as you have workers. The default worker pool is set to 100, so that means you have to make at least that available. You can change that in `config/database.yml` through the `pool` attribute.
--- a/lib/action_cable/engine.rb
+++ b/lib/action_cable/engine.rb
@@ -1,9 +1,14 @@
 require 'rails/engine'
 require 'active_support/ordered_options'
+require 'action_cable/helpers/action_cable_helper'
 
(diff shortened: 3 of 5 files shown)


lib/assets/javascripts/cable.coffee
```diff
@@ -5,4 +5,10 @@
   PING_IDENTIFIER: "_ping"
 
   createConsumer: (url) ->
+    unless url?
+      for tag in document.getElementsByTagName("meta")
```
ascrazy: `document.querySelector("meta[name=websocket-url]")`

lib/assets/javascripts/cable.coffee
```diff
@@ -5,4 +5,9 @@
   PING_IDENTIFIER: "_ping"
 
   createConsumer: (url) ->
+    unless url?
+      tag = document.querySelector("meta[name='websocket-url']")
+      if tag?
+        url = tag.getAttribute("content")
+
```
jeremy: May be condensed to
```coffeescript
url ?= document.querySelector("meta[name='websocket-url']")?.getAttribute("content")
```
and extract the default value
```coffeescript
createConsumer: (url = @defaultUrl()) ->
  …

defaultUrl: ->
  document.querySelector("meta[name='websocket-url']")?.getAttribute("content")
```

lib/action_cable/helpers/websocket_url_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+      #   @App = {}
+      #   App.cable = Cable.createConsumer()
+      #
+      # Make sure to specify the correct server location in each of your environments
+      # config file:
+      #
+      #   config.action_cable.url = "ws://example.com:28080"
+      def websocket_url_meta_tag
+        tag "meta", name: "websocket-url", content: Rails.application.config.action_cable.url
+      end
```
jeremy: Would name these methods to reflect that it's for Action Cable, not a generic "websocket."

jeremy:
Nice approach! We use a meta tag also.

dhh:
👍– can you rebase? We'll get this merged.

smellsblue:
@dhh done. I had to add `url` to `ActionCable::Server::Configuration`, otherwise the `action_cable.set_configs` initializer raises a NoMethodError.

jeremy:
Please squash to a single commit, too, @smellsblue. Thanks!

lib/assets/javascripts/cable.coffee.erb
```diff
@@ -4,5 +4,8 @@
 @Cable =
   INTERNAL: <%= ActionCable::INTERNAL.to_json %>
 
-  createConsumer: (url) ->
+  createConsumer: (url = @defaultUrl()) ->
     new Cable.Consumer url
+
+  defaultUrl: ->
+    document.querySelector("meta[name='action-cable-url']")?.getAttribute("content")
```
javan: Naming here suggests there's a default URL, but really it needs to be configured.

How about:
```coffee
createConsumer: (url = @getConfig("url")) ->
  ...
  
getConfig: (name) ->
  element = document.head.querySelector("meta[name='action-cable-#{name}']")
  element?.getAttribute("content")
```

smellsblue:
@jeremy @javan done :-)

Issue #74, opened 2015-09-05
Redis subscribed race condition


smellsblue:
I have a post action which in turn notifies a thread I have running to stop its processing. When the processing is complete, this in turn triggers some broadcast messages to tell the browser to update the view with some new details. The thread doesn't die right away, so after the redirect from the post, the old state will be visible until the ActionCable broadcast message comes in.

Unfortunately, I am hitting a race condition... the browser reloads from the redirect, but then never gets the message with the new state. At first I thought this might be a race condition from between when the browser first loads to when the ActionCable websocket is actually connected. So, I listened for the consumer to be connected and pulled the state via a JSON request... thus I should (for sure) get the correct state after the redirect from the post:

```coffee
App.myChannel = App.cable.subscriptions.create "MyChannel",
  connected: ->
    # ... call JSON request and update state

  received: (data) ->
    # ... update state
```

This didn't solve the problem. So I dug a bit more and found that my redis subscription wasn't successfully subscribed until after my websocket was connected, and that was the race condition. I have a pull request that I will link to this issue which demonstrates a possible fix such that the following would solve my race condition:

```coffee
App.myChannel = App.cable.subscriptions.create "MyChannel",
  subscribed: ->
    # ... call JSON request and update state

  received: (data) ->
    # ... update state
```

I realize my solution is a bit hacky and probably not the ultimate desired solution, but I would love to dig in and help with the final solution. Where would be the best place to discuss how this should actually be solved?

dhh:
We've seen the same issue. @lifo did you have a take on what we needed to do to solve this?

smellsblue:
@lifo I've been thinking about this problem a bit, and I feel like keeping just the "connected" event would be ideal... It keeps the API simpler. With that, 2 solutions come to my mind:
* Delay the websocket response from the ActionCable server until Redis has responded, so the JavaScript doesn't have to change
* Send a special message to the client once Redis has responded, delaying the "connected" event until this is seen

Do you think either of these might work?

kroehre:
@lifo I've also run in to this issue. Sometimes a background process completes quickly and broadcasts before the browser has connected. I've overcome this by using an intermediate object that is persisted in Redis for a short while when the process completes. It broadcasts immediately, which works when the browser has already connected. In my channel, when a subscription is started, I look up the intermediate object to see if there's a message and transmit it if so. This covers the case when the browser wasn't subscribed when the process finished.

javan:
[...]

PR #75, opened 2015-09-05
Capture when redis subscription is definitely active

smellsblue:
This pull request demonstrates why I am getting a race condition in issue #74. This probably isn't ideal for the solution, but I would love to be involved in the final solution.

A couple thoughts I've had:
* Maybe the client doesn't notify "connected" until the subscription is active?
* This current solution is a problem if you want to send "subscribed" as a message to the client... perhaps the final solution should have no such possible clash?
* Maybe there is a way to wait for the redis subscription before continuing with the socket connection so the connection message is actually connected and subscribed...?
* Maybe there are similar sort of issues when a reconnect needs to happen?

### DIFF
--- a/lib/action_cable/channel/streams.rb
+++ b/lib/action_cable/channel/streams.rb
@@ -72,7 +72,8 @@ module ActionCable
         callback ||= default_stream_callback(broadcasting)
 
         streams << [ broadcasting, callback ]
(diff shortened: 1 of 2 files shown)


dhh:
The connected() callback is now not called until everything is setup and ready to go. So this isn't needed any more.

PR #78, opened 2015-09-16
Allow relative urls for createCable

kbrock:
If you are running the websocket channel in puma along with rails, the following solution can address #72 - This lets you pass in an absolute or relative url, with or without a host/port.

This can be added after the code in #73, if that solution is also desired.

If this looks good, I can spend some time adding support for IE6 and more edge cases. See [gist](https://gist.github.com/jlong/2428561) by @jlong for more information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ Example::Application.routes.draw do
 end
 ```
 
-You can use `App.cable.createConsumer('ws://' + window.location.host + '/websocket')` to connect to the cable server.
+You can use `App.cable.createConsumer('/websocket')` to connect to the cable server.
 
 For every instance of your server you create and for every worker your server spawns, you will also have a new instance of ActionCable, but the use of Redis keeps messages synced across connections.
 
--- a/lib/assets/javascripts/cable.coffee.erb
+++ b/lib/assets/javascripts/cable.coffee.erb
@@ -5,8 +5,19 @@
   INTERNAL: <%= ActionCable::INTERNAL.to_json %>
 
   createConsumer: (url = @getConfig("url")) ->
-    new Cable.Consumer url
+    new Cable.Consumer webSocketURL(url)
 
   getConfig: (name) ->
     element = document.head.querySelector("meta[name='action-cable-#{name}']")
     element?.getAttribute("content")
+
+  webSocketURL = (url) ->
+    if url and not /^wss?:/i.test(url)
+      a = document.createElement("a")
+      a.href = url
+      # in internet explorer, the next line properly populates the protocol
+      a.href = a.href
+      a.protocol = a.protocol.replace("http", "ws")
+      a.href
+    else
+      url


marcinruszkiewicz:
Worth noting that not every browser supports String.prototype.startsWith (current Safari on OSX in particular), so you will most likely want to use some polyfill with that. Quite useful anyway.

kbrock:
@marcinruszkiewicz I can dive in. Just wasn't sure if it was even worth investigating.

marcinruszkiewicz:
@kbrock It's still the default browser on OS X, so I figure it would be nice to have. :)

kbrock:
@marcinruszkiewicz ha. No, I didn't want to dive too deep into solving this for multiple browsers if we didn't even like the idea of auto detecting the url.

marcinruszkiewicz:
Allright. Well, from what the whole gem looks like, I'd hazard a guess that most of the time an app using it would not require a different websocket url anyway (especially that you can deploy it on heroku and so on as one app), so relative url's look to me like a nice thing to have or even a good default behaviour. 

Not the one calling the shots though, so yeah. :)

kbrock:
I'm looking into adding support for all web browsers that support [websockets](http://caniuse.com/#feat=websockets). This problem is really interesting and has been around for a [while](http://tantek.com/2011/238/b1/many-ways-slice-url-name-pieces) - but that means there are some good [answers](http://james.padolsey.com/javascript/parsing-urls-with-the-dom/)

Also good that there are [unit testing services](https://www.jonathanfielding.com/unit-test-your-code-across-multiple-browsers-with-browserstack/) out there to verify.

kbrock:
@marcinruszkiewicz thank you for suggestions.

I created the following [test suite](https://github.com/kbrock/websocket_url). I will hook it up to travis and sauceLabs to test agains a number of websocket compatible browsers like internet explorer.

So far, I tested the tests on a Mac with Chrome, Firefox, and Safari and the tests pass.
(Have not tested the coffee script version with action cable version yet.)

jeremy:
👍

kbrock:
This works in mac: chrome, firefox, safari, and windows: chrome, ie.
I've been having trouble setting up a windows firefox test. (This is the first time I've setup javascript unit tests across multiple browsers using a web service).

You can find the [travis](https://travis-ci.org/kbrock/websocket_url) results. This is only testing the websocket url discovery piece.

If you share your test matrix, then i can verify it across more browsers.

lib/assets/javascripts/cable.coffee
```diff
@@ -5,4 +5,12 @@
   PING_IDENTIFIER: "_ping"
 
   createConsumer: (url) ->
+    if url && (!url.startsWith("wss:") || !url.startsWith("ws:"))
```
javan: String#startsWith isn't widely supported: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility

lib/assets/javascripts/cable.coffee
```diff
@@ -5,4 +5,12 @@
...
+      parser = document.createElement('a')
+      parser.href = url
+      parser.href = parser.href
+      parser.protocol = parser.protocol.replace("http", "ws")
+      url = parser.href
+
     new Cable.Consumer url
+
```
javan: Here's another way to write this without creating an element:

```coffee
createConsumer: (url) ->
  if url = normalizeURL(url)
    new Cable.Consumer url
  else
    throw new Error "Invalid WebSocket URL: '#{url}'"

normalizeURL = (url) ->
  if url and not /^wss?:/i.test(url)
    protocol = location.protocol.replace("http", "ws")
    path = url.replace(/^\//, "")
    "#{protocol}//#{path}"
  else
    url
```

lib/assets/javascripts/cable.coffee
kbrock: thanks so much, updating my [tests](https://github.com/kbrock/websocket_url/blob/master/src/websocket_url.spec.js) to add portion to it.

kbrock:
@javan thank you for your feedback. I've included your feedback ==> kbrock/websocket_url#1

1. Converted test script to coffee.
2. Added some tests around passing in a url with `wss://`.
3. Added `if` check around `wss://` though it was not necessary.
3. There are [conversations](https://gist.github.com/jlong/2428561) around the merits of a dom vs regular expressions for parsing urls. While I like regular expressions, the dom approached seemed simpler to me.
4. The regular expression needs to be more thorough, and I'm not sold it is the right solution.

kayakyakr:
@kbrock I'm not a fan of the document method because of this:

```javascript
parser = document.createElement('a');
parser.href = "www.google.com";
console.log(parser.href); // "https://github.com/rails/actioncable/pull/78/www.google.com"
```

It's cute but hacky.

kbrock:
@kayakyakr thanks I'll add `"//www.google.com"`

I'd like to keep as close to the url syntax as possible. It is relatively well documented. It is not perfect, and not 100% consistent across all browsers/programming languages, but it is pretty good. And boy is it documented along with the common pitfalls. I (and you) saw your example and knew exactly what was wrong. Heck, I'd even say that if you show that bug to any php/asp/html developer and they will tell you how to fix. That kind of documentation/support is so nice.

Having said that, all rails apps are probably going to have one of 3 cases: 1) absolute path, 2) same host but different port, or 3) different host.

Making those 3 cases as easy as possible is preferred. And the url syntax does not currently support `":8080"`, which is the current default answer for rails and websockets/channels.

kbrock:
@kayakyakr @javan I am aiming for simple rather than clever or cure.

In searching the net, there are 2 common ways to convert a relative url to an absolute one. 1) Regular expression or 2) using the dom.

The regular expression solutions looks quite advanced, and the dom solutions looked simpler. Many people I respect did weigh on this like Ilya Grigorik and Paul English, and that led me to the current solution. e.g.: [ref](http://james.padolsey.com/snippets/parsing-urls-with-the-dom/)

More important than the actual implementation, do we at least agree that a url going in and a url coming out is a good route to take?
If so, I can look into implementing a regular expression implementation.

kayakyakr:
After sleeping on it, I'm no longer as opposed to the dom method as a fallback. Using https://developer.mozilla.org/en-US/docs/Web/API/URL/URL would be the best, but it doesn't support ie before 10, unfortunately. Dom behaves like the URL api, which is the best that could be hoped for, I guess.

javan:
Web Sockets aren't supported before IE 10 either (http://caniuse.com/#feat=websockets) so anything less isn't a target browser (currently).

kayakyakr:
Ha, good point @javan 

In that case, instead of using the dom, how about using the URL api which works similarly and is supported in IE10 on up?

javan:
Are you certain `URL` is supported in IE10+? https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#Browser_compatibility

kayakyakr:
Linked the wrong one: https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility

I don't have a windows system with IE to test. It actually looks like it's broken in Edge, returning an error "function expected". Sign, Microsoft is why we can't have nice things.

kbrock:
I need to rewrite the test to split each browser out into a different travis target - after I do that, the results will be consistent and I can see about rewriting this for `URL`
currently testing on ie 10

dhh:
Any progress on this?

kbrock:
@dhh it works fine across safari, chrome, firefox, ie. Just had trouble getting travis to call into saucelabs. (worked fine from my command line)

I haven't rewritten this using `URL` yet, was

javan:
@kbrock, your implementation here differs from https://github.com/kbrock/websocket_url/blob/master/src/websocket_url.coffee, and looks like there's a merge conflict. Mind freshening it up?

kbrock:
I was happy with the `DOM` approach, but once I get my sencha testing environment working, then I can look more into the `URL` implementation.

Not sure if this implementation is good enough for now?

lib/assets/javascripts/cable.coffee
```diff
@@ -0,0 +1,19 @@
...
+  websocketUrl = (url) ->
+    if url and not /^wss?:/i.test(url)
+      a =  document.createElement('a')
+      a.href = url
+      # ie fix for relative urls
+      a.href = a.href
+      a.protocol = a.protocol.replace("http", "ws")
+      a.href
+    else
+      url
```
javan: This file was renamed to `cable.coffee.erb` in master and has been modified since. You'll need to move your changes over.

lib/assets/javascripts/cable.coffee
```diff
@@ -0,0 +1,19 @@
...
+
+@Cable =
+  PING_IDENTIFIER: "_ping"
+
+  createConsumer: (url) ->
+    new Cable.Consumer websocketUrl(url)
+
+  websocketUrl = (url) ->
+    if url and not /^wss?:/i.test(url)
+      a =  document.createElement('a')
```
javan: :scissors: extra space after `=`, and use double quotes (`"a"`) for consistency please.

README.md
```diff
@@ -374,6 +374,16 @@ And finally, create your consumer like so:
...
+If you are using the same server to handling the websockets, you can create the connection with a relaitve path:
+
+```coffeescript
+# app/assets/javascripts/application_cable.coffee
+#= require cable
+
+@App = {}
+App.cable = Cable.createConsumer "/"
+```
+
```
javan: This section of the README describes using `<%= action_cable_meta_tag %>` so I don't think this belongs here. Maybe just add a sentence to the description following the first `Cable.createConsumer` example.

lib/assets/javascripts/cable.coffee
```diff
@@ -0,0 +1,19 @@
...
+      a.href = url
+      # ie fix for relative urls
```
javan: What exactly does this fix?

README.md
```diff
@@ -374,6 +374,16 @@ And finally, create your consumer like so:
 App.cable = Cable.createConsumer()
 ```
 
+If you are using the same server to handling the websockets, you can create the connection with a relaitve path:
```
javan: Typo, should be "relative"

lib/assets/javascripts/cable.coffee
kbrock: the next line looks pointless. `a.href = a.href`
In internet explorer, it fixes relative urls.
I'll change "ie" to "internet explorer"

javan: Fixes them how? What happens without `a.href = a.href`? That's what I'd like to document.

kbrock:
@javan thanks so much for all your help on this.

Let me know what else I can do to make this easier to merge.

FYI I put in a PR for the other half of my install #139

lib/assets/javascripts/cable.coffee.erb
```diff
@@ -5,8 +5,19 @@
...
   getConfig: (name) ->
     element = document.head.querySelector("meta[name='action-cable-#{name}']")
     element?.getAttribute("content")
+
+  websocketUrl = (url) ->
+    if url and not /^wss?:/i.test(url)
+      a = document.createElement("a")
+      a.href = url
+      # in internet explorer, the next line converts a relative url to an absolute one.
+      a.href = a.href
```
javan: I can't see that this actually converts or fixes anything. I tested in IE 10, 11 and Edge. And always got the same results with this test:

```js
window.location.toString()
=> "https://github.com/rails/actioncable"

a = document.createElement("a")

a.href = "/websocket"
a.href
=> "https://github.com/websocket"

a.href = a.href
a.href
=> "https://github.com/websocket"
```

README.md
```diff
@@ -98,6 +98,8 @@ The ws://cable.example.com address must point to your set of Action Cable server
 must share a cookie namespace with the rest of the application (which may live under http://example.com).
 This ensures that the signed cookie will be correctly sent.
 
+If the Action Cable server is running on the rails application, a relative url can be used. It will use the approperiate protocol.
```
javan: Sorry for the runaround, but I think it would be better to remove this and update the `createConsumer` example [here in the README](https://github.com/rails/actioncable/blob/master/README.md#in-app) to `App.cable.createConsumer("/websocket")` instead of piecing together a URL. No need to explain more than that.

lib/assets/javascripts/cable.coffee.erb
```diff
@@ -5,8 +5,19 @@
...
 
   createConsumer: (url = @getConfig("url")) ->
-    new Cable.Consumer url
+    new Cable.Consumer websocketUrl(url)
 
   getConfig: (name) ->
     element = document.head.querySelector("meta[name='action-cable-#{name}']")
     element?.getAttribute("content")
+
+  websocketUrl = (url) ->
```
javan: Function name should be camelCased: `webSocketURL`

kbrock:
@javan sorry - incorrectly reported the internet explorer issue

[problem example](https://gist.github.com/jlong/2428561#gistcomment-1368260) and [solution](https://gist.github.com/jlong/2428561#gistcomment-1368260)

README.md
kbrock: this is great. thanks

javan:
Ok, I see the problem in IE 11:
```js
a = document.createElement("a")
a.href = "/websocket"

a.protocol
=> ""

a.protocol = a.protocol.replace("http", "ws")
=> Invalid argument.

a.href = a.href
a.protocol
=> "https:"
```

Rather than work around it, I think it would be more clear to set the protocol explicitly from `window.location`.
```js
a = document.createElement("a")
a.href = "/websocket"
a.protocol = window.location.protocol.replace("http", "ws")
a.href
=> "wss://github.com/websocket"
```

kbrock:
@javan I think you want to respect the protocol that the user specifies (if they do) So it would look more like:

```javascript
a.protocol = (a.protocol || window.location.protocol).replace("http", "ws")
```

Guess that could work.
I kinda like the ie hack, but can replace with this if you are game

javan:
Went with the original IE fix and cleaned things up a bit. Thanks!

dhh:
We gotta get this ported over to the rails/master branch
merge-action-cable. This repo is essentially dead now.

javan:
I'll open a PR there.

PR #83, opened 2015-10-03
use correct filename in example

### LINKED ISSUE #80: `web_notifications.rb` file from the example should be `web_notificaitons_channel.rb`?
Hi,

I'm actually not sure about this issue, but, when I was trying to create a subscription using a file named as in the example from the README.md:

    # app/channels/web_notifications.rb

the subscription just didn't work.

However, the subscription started working when I changed the file name to:

    # app/channels/web_notifications_channel.rb

I tried changing only the name of the file by removing the "_channel" part, and the subscription stopped working. Putting it back on, then it started working again. But, I'm not sure if ActionCable takes the file name in consideration.

gregmolnar:
closes #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ This is a web notification channel that allows you to trigger client-side web no
 streams:
 
 ```ruby
-# app/channels/web_notifications.rb
+# app/channels/web_notifications_channel.rb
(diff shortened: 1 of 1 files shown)


eileencodes:
Thanks! :+1:

PR #85, opened 2015-10-07
Add support for cross site forgery protection

lifo:


lib/action_cable/connection/base.rb
```diff
@@ -165,6 +165,28 @@ def on_close
...
 
 
+        def allow_request_origin?
+          return true if server.config.disable_request_forgery_protection
+
+          if env['HTTP_ORIGIN'].present?
+            origin_host = URI.parse(env['HTTP_ORIGIN']).host
+
+            allowed = if server.config.allowed_request_origins.present?
+              Array.wrap(server.config.allowed_request_origins).include? origin_host
```
rafaelfranca: Is `allowed_request_origins` to be a hash? If not we can use `Array()` instead of `Array.wrap` and avoid requiring it.

rafaelfranca:
cc @rails/security

test/connection/base_test.rb
```diff
@@ -16,6 +16,7 @@ def disconnect
 
   setup do
     @server = TestServer.new
+    @server.config.disable_request_forgery_protection = true
```
rafaelfranca: Instead of disabling the protection can we make sure the env is always correct?

test/connection/cross_site_forgery_test.rb
```diff
@@ -0,0 +1,55 @@
+require 'test_helper'
+require 'stubs/test_server'
+
+class ActionCable::Connection::CrossSiteForgeryTest < ActiveSupport::TestCase
```
rafaelfranca: Missing a test for these two cases:

1. origin is not present.
2. origin is an invalid URL.

jeremy: The Origin header is meant to be treated, and compared, as an opaque identifier. It will happen to parse as a valid URL, but it's a not a requirement we need to enforce. We just need to do straight equality comparison.

test/connection/base_test.rb
jeremy: +1

lib/action_cable/connection/base.rb
```diff
@@ -165,6 +165,28 @@ def on_close
         end
 
 
+        def allow_request_origin?
+          return true if server.config.disable_request_forgery_protection
```
jeremy: Arguably we should never disable this. Instead we can get a future feature that lets someone do a wildcard Origin whitelist match.

lib/action_cable/connection/base.rb
```diff
@@ -165,6 +165,28 @@ def on_close
...
+
+          if env['HTTP_ORIGIN'].present?
+            origin_host = URI.parse(env['HTTP_ORIGIN']).host
```
jeremy: The Origin header is an opaque identifier, so let's do literal comparisons rather than parsing. Whitelist actual Origin values rather than hosts only.

That's important because the URL scheme is included in the Origin as well as, in most browsers, the port number.

lib/action_cable/connection/base.rb
```diff
@@ -165,6 +165,28 @@ def on_close
...
+
+            allowed = if server.config.allowed_request_origins.present?
+              Array.wrap(server.config.allowed_request_origins).include? origin_host
+            else
+              request.host == origin_host
```
jeremy: This is a sensible default, but needs closer consideration. We don't whitelist the Host either, so a matching Host and Origin could allow someone to bypass our Origin checks. Clearer to always require a whitelist entry.

lib/action_cable/connection/base.rb
```diff
@@ -165,6 +165,28 @@ def on_close
...
+            end
+
+            logger.error("Request origin not allowed: #{env['HTTP_ORIGIN']}") unless allowed
+            allowed
+          else
+            logger.error("Request origin missing.")
+            false
```
jeremy: YAGNI also. A missing Origin header will fail the whitelist check as-is.

lib/action_cable/connection/base.rb
lifo: Fixed in 8275cfb20fda9c89c5b8d7fc17f9f88822fc34d2

test/connection/cross_site_forgery_test.rb
rafaelfranca: :+1:

lib/action_cable/connection/base.rb
lifo: How would the Host and the Origin match for a cross site hijacking though? I think we should have some sort of default at least.

PR #89, opened 2015-10-13
Adding some examples to the README.

kayakyakr:
Since ActionCable is an engine, it can be run in-app in the same way that websocket-rails can be (and I will admit to shamelessly stealing the implementation from websocket rails). Also adding examples of passing params to subscription and rebroadcasting received message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,14 +186,16 @@ class WebNotificationsChannel < ApplicationCable::Channel
 ```
 
 ```coffeescript
-# Somewhere in your app this is called, perhaps from a NewCommentJob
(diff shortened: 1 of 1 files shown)


dhh:
Lovely!

dhh:
We can definitely use your help to flesh out the documentation even more.

kayakyakr:
Sure thing. I'm probably going to be running ActionCable in staging later today and, if all goes well, in production (alpha/beta software) later this week. Be happy to add documentation as I find things.

dhh:
Perfect!

PR #90, opened 2015-10-13
add configuration for allowed_requrest_origins to readme

lsylvester:
I am getting `Request origin not allowed: http://localhost:3000` errors since #85 was merged.

This adds example config to allow requests to be made that pass this protection.

NOTE: I tried putting this into the config/environments/*.rb files but it caused issues with the server.logger being nil. I think that we need a `actioncable/railtie` to manage configuring the server at the correct point of the application load process. Would there be interest in a railtie or should that wait until actionable is being merged into rails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ require ::File.expand_path('../../config/environment',  __FILE__)
 Rails.application.eager_load!
 
 require 'action_cable/process/logging'
-
+ActionCable.server.config.allowed_request_origins = ["http://localhost:3000"]
 run ActionCable.server
 ```
 


kaspth:
I'm 👍 on adding a Railtie. Then we should just allow localhost in development :grin:

dam13n:
@lsylvester for now, what's a good work around for this developing locally? tried all sorts of CORS things, but not having any luck.

kaspth:
@dam13n have you tried what @lsylvester has added to the README? :grin:

dam13n:
haha, my bad! i forgot i was reading a PR and not an issue. thanks @kaspth

kaspth:
@dam13n No problem 👍

kaspth:
Continued in #93.

PR #93, opened 2015-10-14
add railtie and set default allowed_request_origins for development

lsylvester:
Add railtie and set default value for allowed_request_origins to http://localhost:3000 for development.

This might be a bit restrictive for people who develop on a different host (like a *.dev) or different ports, so this should change to "*" once support for wildcards in `allowed_request_origins` is supported.

Alternatively, could set `disable_request_forgery_protection` to development.

As discusses in #90.

/cc @kaspth

### DIFF
--- a/lib/action_cable/railtie.rb
+++ b/lib/action_cable/railtie.rb
@@ -1,0 +1,19 @@
+module ActionCable
+  class Railtie < Rails::Railtie
+    config.action_cable = ActiveSupport::OrderedOptions.new
+
+    initializer "action_cable.logger" do
+      ActiveSupport.on_load(:action_cable) { self.logger ||= ::Rails.logger }
+    end
+
+    initializer "action_cable.set_configs" do |app|
+      options = app.config.action_cable
+
+      options.allowed_request_origins ||= "http://localhost:3000" if ::Rails.env.development?
+
+      ActiveSupport.on_load(:action_cable) do
+        options.each { |k,v| send("#{k}=", v) }
+      end
+    end
+  end
+end
--- a/lib/action_cable.rb
+++ b/lib/action_cable.rb
@@ -17,6 +17,8 @@ require 'em-hiredis'
 require 'redis'
 
 require 'action_cable/engine' if defined?(Rails)
+require 'action_cable/railtie' if defined?(Rails)
+
 require 'action_cable/version'
 
 module ActionCable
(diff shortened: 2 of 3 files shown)


lib/action_cable/railtie.rb
```diff
@@ -0,0 +1,19 @@
...
+    config.action_cable = ActiveSupport::OrderedOptions.new
+
+    initializer "action_cable.logger" do
+      ActiveSupport.on_load(:action_cable) { self.logger ||= ::Rails.logger }
+    end
+
+    initializer "action_cable.set_configs" do |app|
+      options = app.config.action_cable
+
+      options.allowed_request_origins ||= "http://localhost:3000" if ::Rails.env.development?
```
jeremy: Could set the default logger here, too, (`options.logger ||= Rails.logger`) so people can `config.action_cable.logger = …` in their apps.

lsylvester: @jeremy The `config.action_cable.logger = …` currently works because the logger is set on line 6 to the Rails.logger. If this in the same structure that I saw in other rails gems. Is it ok like that?

jeremy: Yep, right. Just noticed that it's identical to setting configs :grin: Def makes sense to mirror the existing Railties.

lib/action_cable/railtie.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      ActiveSupport.on_load(:action_cable) do
+        options.each { |k,v| self.send("#{k}=", v) }
```
kaspth: Could rely on implicit self and omit it here.

Find it curious we use `send` and not `public_send` in all Rails gems, but it's okay I guess.

lib/action_cable/railtie.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+    end
+  end
+end
```
kaspth: Add a newline :grin:

lib/action_cable/railtie.rb
kaspth: It doesn't seem like the server configuration tries to normalize `allowed_request_origins`, so it should be set to an array :smile:

lsylvester: @kaspath When the value is used at https://github.com/rails/actioncable/blob/master/lib/action_cable/connection/base.rb#L178 it is converted into an array.

lsylvester:
@kaspth I have updates as per feedback

lib/action_cable/railtie.rb
kaspth: Nice :+1:

kaspth:
Sweet! Thanks for all the work on Action Cable by the way :smile:

PR #94, opened 2015-10-15
Fix tests and Travis builds

lsylvester:
Add connection method to the WorkerTest::Receiver, the connection is being read from the receiver in Worker#run_periodic_timer and this is causing the tests to fail.

Force EM.next_tick to execute with `EM.run_deferred_callbacks`

Currently the tests are passing for me, but the build is still failing on travis with `NoMethodError: undefined method `worker_pool' for #<TestServer:0x00000003bf7f68>`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -104,4 +104,4 @@ DEPENDENCIES
   rake
 
 BUNDLED WITH
-   1.10.5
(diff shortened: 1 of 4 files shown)


lsylvester:
Looks like the tests are passing again

PR #37, opened 2015-07-24
Use Celluloid 0.16.0 until termination issue in 0.17.0 is resolved

javan:
The issue: https://github.com/celluloid/celluloid/issues/637

The issue in action:
```
Loading development environment (Rails 5.0.0.alpha)
>> exit
D, [2015-07-24T12:01:01.490817 #33479] DEBUG -- : Terminating 5 actors...
E, [2015-07-24T12:01:11.494161 #33479] ERROR -- : Couldn't cleanly terminate all actors in 10 seconds!
```

Which causes a 10 second hang before exiting.

PR #95, opened 2015-10-15
Re upgrading Celluloid to 0.17.2

kayakyakr:
As downgraded in: https://github.com/rails/actioncable/pull/37, Celluloid 0.17.2 seems to fix the issues that it has with Spring. This means ActionCable is free to move forward on Celluloid 0.17.2. Example showing the issue is fixed:

```shell
time rails r 'puts "Celluloid @ #{Celluloid::VERSION}, Spring @ #{defined?(Spring) ? Spring::VERSION : "-"}"'
Celluloid @ 0.17.2, Spring @ 1.4.0

real	0m0.654s
user	0m0.408s
sys	0m0.036s
```

```shell
time rails r 'puts "Celluloid @ #{Celluloid::VERSION}, Spring @ #{defined?(Spring) ? Spring::VERSION : "-"}"'
I, [2015-10-14T21:41:51.180579 #24031]  INFO -- : Celluloid 0.17.2 is running in BACKPORTED mode. [ http://git.io/vJf3J ]
Celluloid @ 0.17.2, Spring @ -

real	0m1.967s
user	0m1.670s
sys	0m0.168s
```

There are currently 4 failing tests and 2 errors in the test suite that are failing for the master branch. The test suite also seems to deadlock when run with certain seeds (4042 was one), but I do not believe that to have any relation to celluloid's version.

### DIFF
--- a/actioncable.gemspec
+++ b/actioncable.gemspec
@@ -18,8 +18,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'actionpack',       '>= 4.2.0'
   s.add_dependency 'faye-websocket',   '~> 0.10.0'
   s.add_dependency 'websocket-driver', '~> 0.6.1'
-  # Use 0.16.0 until https://github.com/celluloid/celluloid/issues/637 is resolved
-  s.add_dependency 'celluloid',        '~> 0.16.0'
+  s.add_dependency 'celluloid',        '~> 0.17', '>= 0.17.2'
   s.add_dependency 'em-hiredis',       '~> 0.3.0'
   s.add_dependency 'redis',            '~> 3.0'
   s.add_dependency 'coffee-rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,7 +48,7 @@ PATH
     actioncable (0.0.3)
       actionpack (>= 4.2.0)
       activesupport (>= 4.2.0)
-      celluloid (~> 0.16.0)
+      celluloid (~> 0.17, >= 0.17.2)
       coffee-rails
       em-hiredis (~> 0.3.0)
       faye-websocket (~> 0.10.0)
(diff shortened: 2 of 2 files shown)


kayakyakr:
Bah, tried to get cute by rebasing from the current master (so tests would pass) and it screwed up the pull request mightily.

kayakyakr:
There we go, much better. Working, with tests passing, on Celluloid 0.17.2.

antifun:
:+1: We'll be running off of this since we need to upgrade other celluloid-dependent things. Thanks! (I was about to submit the same PR but happened to notice your circle build and was like, "hey...!")

kayakyakr:
Want me to rebase real quick?

antifun:
Sure.

kayakyakr:
Phew, that was a big merge. With the intervening changes, the `require 'celluloid'` in `lib/action_cable.rb` was moved into specific files. I didn't update those, as it will still work, but down the road, those should be changed to `require 'celluloid/current'`.

antifun:
Thanks for the rebase. For us, and likely many others, the main other celluloid dependency is from Sidekiq, but it looks like that won't be an issue for long as @mperham will be removing it in Sidekiq 4. But until then, we run on the bleeding edge :grin:

badlamer:
+1

actioncable.gemspec
```diff
@@ -18,8 +18,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'actionpack',       '>= 4.2.0'
   s.add_dependency 'faye-websocket',   '~> 0.10.0'
   s.add_dependency 'websocket-driver', '~> 0.6.1'
-  # Use 0.16.0 until https://github.com/celluloid/celluloid/issues/637 is resolved
-  s.add_dependency 'celluloid',        '~> 0.16.0'
```
rafaelfranca: Is there any reason to not allow 0.16.0?

kayakyakr: Because we can't say `0.16.0` OR `> 0.17.2`

We could say `> 0.16.0` and then add a note to the readme that `0.17.0` and `0.17.1` have breaking issues with Spring.

The other downside to supporting both `0.16` and `0.17` is that the primary require for celluloid has moved from `require 'celluloid'` to `require 'celluloid/current'` which would mean we'd either need to force celluloid to run in what they term "fallback" mode and put up with the warnings, or put version switching code in the many places celluloid is required.

Issue #92, opened 2015-10-14
Lack of Generators


TheAustinSeven:
As of right now there are no generators for new actioncable connections. It seems that to adhere to the  "rails-y" way of things there should be at least an initializing generator that adds the code necessary to start adding connections. Is this an issue or is there a reason this hasn't been done yet?

kaspth:
Action Cable is alpha software, so it hasn't been done yet. Here's a pull request adding channel generators to take inspiration from: https://github.com/rails/actioncable/pull/17 — in case you want to try your hand at a pull request :grin:

TheAustinSeven:
Awesome, I'll get right on it.

seuros:
Generators should be added after the API is stable. I will update the PR when we reach that state.
Adding generators now will just make refactoring the API more time consuming.

TheAustinSeven:
@seuros There are still some things that should be added. I started working on one that has some aspects more similar to the generators we see with other parts of rails.

TheAustinSeven:
The pull request I submitted is #96.

PR #96, opened 2015-10-15
Adding generators for actioncable

TheAustinSeven:
This is certainly inspired by #17. 

This pull request adds generators for not only the server side files necessary, but also some Coffeescript files that have the code necessary to guide someone new to Actioncable as to how to set up their client side code. This code is commented out by default. 

This pull request also improves on #17 by ensuring that some of the necessary files (channel.rb and connection.rb), are created and have some starting code.


lib/rails/generators/channel/templates/application_cable_connection.rb.erb
```diff
@@ -0,0 +1,19 @@
+module ApplicationCable
+  class Connection < ActionCable::Connection::Base
+    identified_by :current_user
```
kayakyakr: Shouldn't have contents to this class, this will vary from implementation to implementation.

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails/generators/named_base'
+
+module Rails
+  module Generators
+    class ChannelGenerator < Rails::Generators::NamedBase
+      desc 'This generator creates the Ruby and Coffeescript files to support Actioncable'
```
jeremy: "CoffeeScript"

"Action Cable"

kayakyakr:
Probably not the place for this discussion, but should the empty ApplicationCable::Connection and ApplicationCable::Channel be defined within the gem itself? This would prevent the need for the two files to be generated unless someone needs to override them.

lib/rails/generators/channel/channel_generator.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+        unless File.exist?('app/channels/application_cable/connection.rb')
+          template 'application_cable_connection.rb.erb', File.join('app/channels/application_cable', class_path, "connection.rb")
+        end
+        unless File.exist?('app/assets/javascripts/channels')
+          template 'index.coffee.erb', File.join('app/assets/javascripts/channels', "index.coffee")
+        end
+        unless File.read("app/assets/javascripts/application.js").include?("//= require channels")
+          inject_into_file "app/assets/javascripts/application.js", "//= require channels\n", :before => "//= require_tree .\n"
+        end
```
jeremy: Assumes a lot about the structure of `application.js`. Suggest not attempting to modify this.

TheAustinSeven: Agreed, I just looked over some other gems, and none of them generate them in this way.

lib/rails/generators/channel/templates/channel.coffee.erb
```diff
@@ -0,0 +1,8 @@
+# App.<%= channel_name %> = App.cable.subscriptions.create "<%= class_name %>",
+  # collection: -> $("[data-channel='<%= channel_name %>']")
+
+  # connected: ->
+    # Called once the subscription has been successfully completed
+
+  # received: (data) ->
```
jeremy: `connected` and `received` probably needn't be commented, huh? OK to be no-ops, and clearer to start with working code.

lib/rails/generators/channel/templates/channel.rb.erb
```diff
@@ -0,0 +1,11 @@
+<% module_namespacing do -%>
+class <%= channel_name %> < ApplicationCable::Channel
+  def subscribed
+    #Performed when a websocket is opened
+  end
+
+  def unsubscribed
+    #Performed when a websocket is closed
```
jeremy: Space after `# ` to match generated code style.

lib/rails/generators/channel/templates/channel.coffee.erb
```diff
@@ -0,0 +1,8 @@
+# App.<%= channel_name %> = App.cable.subscriptions.create "<%= class_name %>",
```
jeremy: Ditto, uncomment

lib/rails/generators/channel/templates/index.coffee.erb
```diff
@@ -0,0 +1,9 @@
+#= require cable
+#= require_self
+#= require_tree .
+
+# Uncomment out this code for a basic implementation of actioncable
+# @App ?= {}
+
+# Define the location of your websockets here
+# App.cable = Cable.createConsumer 'ws://localhost:28080'
```
jeremy: Perhaps use the application name as the toplevel object instead of `App`. So we'd have `@MyApp` here and could uncomment.

lib/rails/generators/channel/templates/channel.coffee.erb
kayakyakr: What do you think about namespacing with the app name? Instead of `App.<%= channel_name %>` you could use `<%= Rails.application.class.parent.to_s. %>.<%= channel_name %>`.

I'm not sure what others think about that, though. I prefer to namespace based on my own app's name, but dunno how often that's used.

lib/rails/generators/channel/templates/index.coffee.erb
kayakyakr: Ha, guess I wasn't the only one thinking the same thing, @jeremy

kayakyakr:
I had a thought: there should probably be a `action_cable:install` generator to create: `app/channels/`, `app/channels/application_cable/connection`, `app/channels/application_cable/channel`, `config/initializers/action_cable.rb`, `config/redis/cable.yml`, and, if we want to make the in-app the standard way (for now) to run the cable server, add in the `/websocket` route.

Thoughts?

kaspth:
The application channel and connection should be generated in apps, and not here. Saves the confusion of what files are needed if suddenly you do need customization.

👍 on having an install generator to get people quickly set up. Don't think we should add an initializer, as Rails doesn't have initializers scoped at the library level, but at the feature level (e.g. cookie serializer and assets don't go in an `action_pack.rb` initializer).

Not sure if the `/websocket` route is what we want. Seems like it's exposing Action Cable innards and not keeping with the rest of our terminology. How about `/<%= application_name %>_cable`?

kayakyakr:
@kaspth on your first point, I agree with you, but right now those files are required to run Action Cable. If default versions of those classes are provided inside Action Cable and are overridden to add customization, then they don't need to be generated at all.

As to your final point, `/websocket` is just a placeholder until official terminology is developed. I like `/<%= application_name %>_cable`.

seuros:
@kayakyakr That the reason i told you it too early to have generator merged.

kayakyakr:
You told me that? You talking about on Reddit? Cause that was a conversation about Celluloid, not generators.

kaspth:
@seuros doesn't mean we can't work on them :smile:

@kayakyakr if they are required then it makes even more sense to have them in the application from the get go. I don't find having them in Action Cable that much of a convenience for users. Besides how would they work with Rails' autoloading? If we've already defined `ApplicationCable::Channel`, the one users defined will never be loaded and is confusing.

The nice thing about generators is that they spur you into more learning: "Oh, what does `app/channels/application_cable/channel` do?"

I think having those files defined in Action Cable would be a poor user experience and any convenience minimal and short lived. What do you think? :smile:

kaspth:
@kayakyakr I think @seuros was talking about here: https://github.com/rails/actioncable/issues/92#issuecomment-148103952. :smile:

@seruos About the API being stable or not, I don't think there's a huge chance of it changing now. Since @jeremy has already reviewed I take it he thinks it's okay (do correct me if you disagree @jeremy :smile:).

kayakyakr:
Ah, @seuros had the wrong tag. That's why I was confused.

Re default: I think that Rails will load both with the one in Action Cable loaded first (on application load) and then the one in the application, but it's been a while since I've played around with rails autoloading. Maybe I'll try it to see if it works.

In my production Action Cable app (yes, I'm crazy, using Action Cable in production, but it's working much, much better than websocket-rails), those two files are empty, just the default implementation. I'd be curious to know how many others will implement without ever needing to modify those?

The other option is to fix `/lib/action_cable/server/configuration.rb:17:in 'initialize': uninitialized constant ActionCable::Server::Configuration::ApplicationCable (NameError)` which I think has to do with the missing `ApplicationCable::Connection` and ensure channels can descend from `ActionCable::Channel::Base` directly. The install generator can still generate those files, but they wouldn't be required.

TheAustinSeven:
With this latest commit there are now two generators: `action_cable:install` and `channel`

seuros:
@kaspth You are right, my bad.
@kayakyakr Sorry, i wanted to take the PR author. :)

lib/generators/install/templates/cable.yml.erb
```diff
@@ -0,0 +1,14 @@
+# production: &production
+#   :url: redis://localhost:6379
+#   :host: localhost
+#   :port: 6379
+#   :timeout: 1
+
+local: &local
```
kaspth: Would just call this development and have test extend that :grin:

kaspth:
@TheAustinSeven Unless @seuros wants to cede the channel generators to you, I'd say he should get to finish that at: https://github.com/rails/actioncable/pull/17.

Especially because there's already hefty discussion about those, so let's keep this more focused :heart:

TheAustinSeven:
I'm fine with that. It seems reasonable since he started the discussion and the code in #17. We definitely have a good starting point here.

kaspth:
@kayakyakr @seuros was pretty confused as well :grin:

@kayakyakr Rails' autoloading hooks via `const_missing`, so if the `ApplicationCable::Connection` constant is missing, the autoload paths are searched and the first `application_cable/connection.rb` file found is loaded. So Action Cable would just need to be after `app/channels` to make it work.

Here's another reason why I don't think it's a good idea. The namespace is `ApplicationCable` which means it's tailored for your application (even if there's no customization other than a subclass), Action Cable has no business defining that.

I'd look to close this part of the discussion now. I haven't seen an argument from you that's swayed me to change my mind. How do you feel about that? :heart:

kaspth:
@TheAustinSeven sweet! Yeah, I think this is looking good.

kayakyakr:
That's fine @kaspth, though what do you think about fixing the crash when `ApplicationCable::Connection` is not defined?

lib/generators/install/install_generator.rb
```diff
@@ -0,0 +1,39 @@
...
+          template 'application_cable_channel.rb.erb', File.join('app/channels/application_cable', class_path, "channel.rb")
+        end
+        unless File.exist?('app/channels/application_cable/connection.rb')
+          template 'application_cable_connection.rb.erb', File.join('app/channels/application_cable', class_path, "connection.rb")
+        end
+        unless File.exist?('app/assets/javascripts/channels/index.coffee')
+          template 'index.coffee.erb', File.join('app/assets/javascripts/channels', "index.coffee")
+        end
+        unless File.exist?('config/redis/cable.yml')
+          template 'cable.yml.erb', File.join('config/redis', class_path, "cable.yml")
```
kaspth: The final argument to all these `template` calls use double quoted strings, let's change them to single quotes to match all the other instances here :smile:

kaspth:
I'd say we should check that `ApplicationCable::Connection` is defined. If it isn't we tell people that `ApplicationCable::Connection` is missing and they can add it by running `action_cable:install` :grin:

(@TheAustinSeven here's where the `defined?` check should be added: https://github.com/rails/actioncable/blob/master/lib/action_cable/server/configuration.rb#L17)

TheAustinSeven:
@kaspth should we throw an error to exit or is there some graceful exit that we should employ instead.

kaspth:
Right now it raises a `NameError` and exists, we just need to improve the message.

```ruby
unless defined?(ApplicationCable::Connection)
  raise NameError, "You haven't defined ApplicationCable::Connection, run action_cable:install to have it setup for you."
end
```

Something like this could do :grin:

TheAustinSeven:
Awesome, that is almost exactly what I had. They will still get a different error if they create a channel that is a subclass of connection, but I don't know if that is something that we can address.

kaspth:
Great! No, that sounds too much like an edge case to me. Let's not bother :grin:

lib/action_cable/server/configuration.rb
```diff
@@ -53,4 +57,3 @@ def config_for(path)
     end
   end
 end
-
```
kaspth: Keep this line.

lib/generators/install/templates/cable.yml.erb
```diff
@@ -0,0 +1,14 @@
...
+#   :host: localhost
+#   :port: 6379
+#   :timeout: 1
+
+development: &development
+  :url: redis://localhost:6379
+  :host: localhost
+  :port: 6379
+  :timeout: 1
+  :inline: true
```
kaspth: Do these need to be symbol keys?

kaspth:
Can you squash your commits? :grin:

lib/generators/install/templates/cable.yml.erb
kayakyakr: It's a direct copy from the README, though the configuration section (and installation instructions in general) in the README will need to be rewritten after this commit.

lib/generators/install/templates/index.coffee.erb
```diff
@@ -0,0 +1,8 @@
+#= require cable
+#= require_self
+#= require_tree .
+
+@<%= application_name %> ?= {}
```
javan: Perhaps we can set aside some dedicated objects on `Cable` for storing the consumer and subscription instances instead of creating another (potentially conflicting) global.

Something like:
```coffee
Cable.consumer = Cable.createConsumer(...)
```
and
```coffee
Cable.subscriptions["<%= channel_name %>"] = Cable.consumer.subscriptions.create(...)
```

kayakyakr: You can do that anyway, if you choose. Yay javascript. But `AppName.property` is a pretty common style for storing persistent objects in JS-heavy apps.

javan: Right, so these generators could potentially clobber their existing `AppName` properties.

kayakyakr: If that's the track you take, then you might as well not generate any JS in the first place.

javan: I'm not arguing against generating JavaScript, just trying to limit our footprint on `window`. I guess I'm not clear what your objection to using `Cable` is. Do you see it being more problematic?

kayakyakr: Just concerned with people confusing that as a requirement, as in Cable requires the consumer be set in order to function. It also gives more of the impression that the recommended way to access is via a global property on Cable rather than by using something appropriate to your own JS framework.

That this is only generated when actioncable is installed, I don't think there is much risk of squashing existing JS, so its main purpose should be in leading newer developers in the right direction.

lib/generators/install/templates/index.coffee.erb
```diff
@@ -0,0 +1,8 @@
+#= require cable
+#= require_self
+#= require_tree .
+
+@<%= application_name %> ?= {}
+
+# Define the location of your websockets here
+<%= application_name %>.cable = Cable.createConsumer 'ws://localhost:28080'
```
kayakyakr: The connection string should be wss-aware: `(window.location.protocol == 'https:' ? 'wss://' : 'ws://')` or something similar. This has caught me out on two deploys.

lib/generators/install/templates/index.coffee.erb
javan: Fair point. I don't think any frameworks recommend setting arbitrary property names on your global app object though so minimally we should namespace them (`AppName.subscriptions.appearance` instead of `AppName.appearance`).

Also, we can make storing your subscriptions on `Cable` the recommended way if we want. The API isn't final, and it might make your code more obvious to other readers. If I came across `Cable.subscriptions.appearance.doSomething()` in a non-channel file, I'd be pretty darn sure it was Action Cable related. `AppName.appearance.doSomething()`, not so sure.

Finally, if we're trying to lead newer developers in the right direction, we shouldn't make subscriptions global at all.

This example in the README
```coffee
App.appearance = App.cable.subscriptions.create "AppearanceChannel",
  appear: ->
    @perform 'appear', ...

$(document).on 'page:change', ->
  App.appearance.appear()
```
could be rewritten as:
```coffee
appearance = App.cable.subscriptions.create "AppearanceChannel",
  appear: ->
    @perform 'appear', ...

$(document).on 'page:change', ->
  appearance.appear()
```

And app makers can decide to expose them as globals if/when they need access from outside.

kayakyakr: I like that last idea better. Not a fan of the idea of global subscriptions hanging off Cable being the default api because to me, if I saw that in code, I'd think, "oh hey, there's an undocumented appearance class in ActionCable, let me go dig through the gem to find it."

And to further your point that code should be rewritten as:

```coffee
cable = Cable.createConsumer "xs://localhost:3000/app_cable"

appearance = cable.subscriptions.create "AppearanceChannel",
  appear: ->
    @perform 'appear', ...

$(document).on 'page:change', ->
  appearance.appear()
```

So should we create a cable.coffee instead of index.coffee and add channel subscriptions that are generated to the end of cable.coffee (if it exists) instead?

lib/generators/install/templates/index.coffee.erb
kayakyakr: If we decide to go with in-app as the standard from #103, this line should be:

```coffee
<%= application_name %>.cable = Cable.createConsumer "#{if window.location.protocol == 'https:' then 'wss' else 'ws'}://#{window.location.host}/<%= application_name %>_cable"
```

TheAustinSeven:
I'll squash once we decide this work is merge worthy. That way I only have to squash once.

lib/generators/install/templates/index.coffee.erb
javan: I'd prefer to generate a new file for each channel subscription so we're going to need at least one global for the consumer instance. If we use `Cable` just for that (`Cable.consumer = Cable.createConsumer(...)`) then we don't have to think about the user's global `AppName` object at all and free ourselves from potential conflicts, naming assumptions, and imposing more globals in their app.

So, we'd generate:
```coffee
# app/assets/javascripts/cable/index.coffee

#= require cable
#= require_self
#= require_tree .

Cable.consumer = Cable.createConsumer(...)
```
and
```coffee
# app/assets/javascripts/cable/subscriptions/appearance.coffee
appearance = Cable.consumer.subscriptions.create "AppearanceChannel",
  ...
```

kayakyakr: What if `Cable.createConsumer` kept track of the cables you've set up? So if you try to connect to an already connected cable, it will return the existing connection instead? Should be fairly easy to key based on the url.

Can you forsee a situation when you would want multiple consumers for the same url?

javan: Dig that idea. 👍

> Can you forsee a situation when you would want multiple consumers for the same url? 

Nope.

javan: Thinking for 5 more seconds.. That means we'd have to put the WebSocket URL in every subscription file, which makes updating it a bit of a chore.

kayakyakr: Not too bad since the url in 90% of cases will probably be generated from window.location

If we expose consumers as an array (though it'd be stored as an object), then you could have your cake.

```coffee
# app/assets/javascripts/cable/index.coffee
Cable.createConsumer(...)
```

```coffee
# app/assets/javascripts/cable/subscriptions/appearance.coffee
appearance = Cable.consumer[0].subscriptions.create "AppearanceChannel",
  ...
```

That is not pretty... But I'm OK with each generated channel looking like:

```coffee
# app/assets/javascripts/cable/subscriptions/appearance.coffee
#= require cable

cable = Cable.createConsumer "#{if window.location.protocol == 'https:' then 'wss' else 'ws'}://#{window.location.host}/app_cable"

appearance = cable.subscriptions.create "AppearanceChannel",
  ...
```

The other option is to always store a single consumer on Cable, disconnect it if you try to create a new consumer via the convenience method, and force users who want to connect to multiple sockets to directly use `new Cable.Consumer(...)`

lib/generators/install/templates/index.coffee.erb
kaspth: When would the protocol ever be anything but `was` or `ws`? Seems like `createConsumer` should just set that up:

```CoffeeScript
<%= application_name %>.cable = Cable.createConsumer 'localhost:28080'
```

Perhaps a `protocol: true` argument would make it clearer that it's automatically handled. Don't know if better documentation is enough.

lib/generators/install/templates/index.coffee.erb
kaspth: Don't think every file should start with that configuration. We should only bend when we have to, so that config should only be output if it's different from the standard config. If not, it should be output in some standard setup.

Not liking `consumer` (singular) being an array either :smile:

Do like @javan's last suggestion :+1: 

Enjoyed the discussion here greatly :heart:

lib/generators/install/templates/index.coffee.erb
kayakyakr: Agreed. That's something that can nearly always be inferred from the host protocol as browsers won't allow mixed content.

It shouldn't be too hard to determine if `xss://` or `xs://` has already been included using a regex, like `/^xss?:\/\//.test(url)`. Wouldn't need an option at all at that point.

lib/generators/install/templates/cable.yml.erb
kaspth: @kayakyakr nice point, @TheAustinSeven are you up for amending the README here too? :grin:

lib/generators/install/templates/index.coffee.erb
kaspth: Agreed. If people are inserting the protocol themselves, we should respect that. Nice regex suggestion :+1:

lib/generators/install/templates/cable.yml.erb
TheAustinSeven: Yep, I'll get right on it!

matthewd:
Sorry I missed this discussion earlier.. @kaspth, isn't the install generator named `rake rails:update`? Do we need a separate one? Did we have one for AJ?

kaspth:
@matthewd Good point, the generators don't make much sense outside of a Rails context. We should make this tie into both `update` and `new` too :smile:

Don't remember the generator situation for Active Job at the time of release, will look into it later.

seuros:
AJ generator generate just jobs. 
I will fix #17 tonight to match with the other generators.

kaspth:
@seuros nice! :grin:

kaspth:
@seuros nice! :grin:

TheAustinSeven:
Yeah, it looks like we may have to wait for the merge into Rails to have the install generator, since that should be in the rails `update` and `new` code and not here. I'll finish this up so that we have a good reference, but I think that @kaspth is right in saying that this should tie in with 'update' and 'new'

dhh:
We are about to merge to rails/master. Let's open this there so we can do the work directly in Rails.

PR #98, opened 2015-10-17
Send subscription confirmation from server to the client to avoid race conditions

lifo:
Without this, it's very easy to send messages over a subscription even before the
redis pubsub has been fully initialized.

Now we delay calling the subscription#connected method on the client side until we
receive a subscription confirmation message from the server.

### DIFF
--- a/lib/action_cable/channel/base.rb
+++ b/lib/action_cable/channel/base.rb
@@ -73,6 +73,8 @@ module ActionCable
       include Naming
       include Broadcasting
 
+      SUBSCRIPTION_CONFIRMATION_INTERNAL_MESSAGE = 'confirm_subscription'
+
       on_subscribe   :subscribed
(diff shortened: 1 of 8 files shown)


dhh:
Seems good to me.

lib/action_cable/channel/base.rb
```diff
@@ -73,6 +73,8 @@ class Base
       include Naming
       include Broadcasting
 
+      SUBSCRIPTION_CONFIRMATION_INTERNAL_MESSAGE = 'confirm_subscription'
```
kaspth: Wouldn't `'confirm_subscription'.freeze` inline be faster, clearer and still reap reduced allocations? :grin:

lifo: I'll just add `.freeze` here.

PR #102, opened 2015-10-21
Allow rejecting subscriptions from the channel

lifo:


kayakyakr:
This is awesome. Was thinking that this was missing.

rafaelfranca:
The feature looks good to me but it lacks documentation.

lifo:
It's WIP. I'm not done yet 😊

rafaelfranca:
:+1:

lib/action_cable/channel/base.rb
```diff
@@ -183,6 +198,14 @@ def subscription_confirmation_sent?
           @subscription_confirmation_sent
         end
 
+        def reject!
```
dhh: IMO, we should just use `#reject`. There's not a non-bang version of this method.

lifo: Yeah, makes sense. Changed in 37fe48928e4d4b254791eb8034c20baabe06a483

Issue #110, opened 2015-10-29
Client disconnect with ActionCable.server.remote_connections


nateless:
I think we need to tell client that he has been disconnected, e.g. we can listen to internal channel and process that event. At the moment server gets the message and closes WS connection. Client starts to reconnect automatically. It would be nice to process this event on client and prevent it or ask user whether he wants to reconnect or no.

javan:
Moving discussion to #112.

PR #112, opened 2015-11-01
close connection from server and prevent from re-connect

nateless:
Regarding #110


dhh:
We definitely need this! Thanks for working on it. Do you think you could see if you could take a stab at some tsts?

nateless:
I will try to make it and update the code, however I'm on vacations till 18 Dec.

Issue #111, opened 2015-10-31
There was an exception - NoMethodError(undefined method `_run_unsubscribe_callbacks' for #<MyChannel>)


dballona:
<img width="1412" alt="screen shot 2015-10-31 at 5 34 07 pm" src="https://cloud.githubusercontent.com/assets/1319459/10865554/b954d93e-7ff5-11e5-806d-cc2fd9cf3fe8.png">

Not sure, but is this where maybe it's here where it was introduced?
https://github.com/rails/actioncable/commit/9f1938d9c10b4892d939374b49362e726023b5d7

 cc @mieko @rafaelfranca 
I'll be happy to solve it if you provide some directions, and if it's the case.

mieko:
@dballona There does appear to have been a merge issue when the callbacks branch was merged (compare [unsubscribe_from_channel](https://github.com/mieko/actioncable/commit/cf426a7ee680e8cd30a4b5afccf7e140537836f4#diff-bc1f6d9f78ab7b4b491641f39548e0c1L140) in the callbacks branch to [what was merged](https://github.com/rails/actioncable/commit/9f1938d9c10b4892d939374b49362e726023b5d7?diff=unified#diff-bc1f6d9f78ab7b4b491641f39548e0c1L148)).  Not sure how to proceed here, as I doubt the branch is still conflict-free.

dballona:
I'll revert your callbacks as they were, do some manual tests and update this issue referencing a PR. cc @mieko

jeremy:
Fixed by #113

PR #113, opened 2015-11-04
Fixing subscription callbacks

dballona:
As per #111 

cc @mieko


mieko:
Thanks @dballona, @jeremy

PR #115, opened 2015-11-05
Fix an error when using multiple gid identifiers

habermann24:
When using multiple ``identified_by`` identifiers that respond to ``to_global_id``, you get a comparison error.  This is because we call ``.sort`` on the array.

This change fixes the issue since strings can always be sorted properly.

### DIFF
--- a/lib/action_cable/connection/identification.rb
+++ b/lib/action_cable/connection/identification.rb
@@ -35,7 +35,7 @@ module ActionCable
         def connection_gid(ids)
           ids.map do |o|
             if o.respond_to? :to_global_id
-              o.to_global_id
+              o.to_global_id.to_s
             else
               o.to_s
             end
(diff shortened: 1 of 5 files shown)


jeremy:
Thanks @habermann24 – could you add a test to demonstrate the expected behavior, too?

habermann24:
OK, i hope i did this in a correct way.

I added a GlobalID stub and a test for multiple identifiers.
With this in place we also have a failing test (without my fix).

You would get:
```
ArgumentError: comparison of GlobalID with GlobalID failed
```

lib/action_cable/connection/identification.rb
```diff
@@ -35,7 +35,7 @@ def connection_identifier
         def connection_gid(ids)
           ids.map do |o|
             if o.respond_to? :to_global_id
-              o.to_global_id
+              o.to_global_id.to_s
```
lsylvester: Maybe we should be using`to_gid_param` instead of `to_global_id.to_s` here, like we do in https://github.com/rails/actioncable/blob/master/lib/action_cable/channel/broadcasting.rb#L20

habermann24:
I changed it to use `to_gid_param` as suggested.

habermann24:
Hey @jeremy, any further suggestions?

PR #117, opened 2015-11-06
Share internal identifiers and message types with the JavaScript client

javan:
Move internal messages to `ActionCable::INTERNAL` and sync  `window.Cable` so we don't have to maintain them in two places.

cc: @lifo


lifo:
Looks great to me!

Issue #99, opened 2015-10-19
Redis-sentinel support


badlamer:

seuros:
Redis sentinels are supported by default in the redis gem.

badlamer:
@seuros but actioncable use em-hiredis gem, which does not support sentinel connections

seuros:
The support should be added in em-hiredis then.

badlamer:
Need a special connection url em-hiredis - **SUBSCRIBE** since the operation can be performed in a mode readonly, it can be used for connection em-hiredis to localhost.
Meanwhile broadcating goes through the driver redis-rb, who once supported the sentinel hook up, but the url's in such a case is redis://sentinel_group - which is not compatible with the em-hiredis.

I`m solved this with monkey patch:

```ruby
module ActionCable
  module Server
    class Base
      def redis
        @redis ||= EM::Hiredis.connect(config.redis[:hiredis_url]).tap do |redis|
          redis.on(:reconnect_failed) do
            logger.info '[ActionCable] Redis reconnect failed.'
          end
        end
      end
    end
  end
end
```
But still needed a scalable solution.

PR #124, opened 2015-11-16
Point Broadcasting Redis to Event Machine Redis

rushingfitness:
This one took me a while to figure out why it was posting to the front end. It seems that calling Redis by itself doesn't call the callback set in streams and instead needs the EM connection.

1.) Switch to EM Redis Instance already established in `ActionCable::Server::Base`
2.) Put in a catch in case EM isn't started yet

### DIFF
--- a/lib/action_cable/server/base.rb
+++ b/lib/action_cable/server/base.rb
@@ -61,6 +61,16 @@ module ActionCable
             # @connections.map &:close
           end
         end
+      rescue RuntimeError
+        ensure_em
+        retry unless (tries -= 1).zero?
+      end
+
+      # Very Rarely EM might die so needs to have a thread established to turn it back on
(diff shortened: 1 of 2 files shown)


badlamer:
Event Machine Redis not support sentinel connections - not scalability.
https://github.com/rails/actioncable/issues/99

rushingfitness:
https://github.com/livehelpnow/em-hiredis-sentinel

@badlamer Perhaps this gem could be merged into AC, or someone could PR it into em-hiredis?

badlamer:
@rushingfitness AC not working with this gem. Broadcasting better done through a redis-rb client.

rushingfitness:
Well if I go with the master branch, broadcasting doesn't even work at all with the redis-rb client.

PR #130, opened 2015-12-05
Allow regexp for a allowed_request_origins array

adamliesko:
Adding the feature of additional regexp check for `allowed_origin_requests`instead of only strict equal comparation.

Fixes #130

### DIFF
--- a/lib/action_cable/connection/base.rb
+++ b/lib/action_cable/connection/base.rb
@@ -180,11 +180,6 @@ module ActionCable
           end
         end
 
-        def allowed_origins_match? origin
-          allowed_origins = Array(server.config.allowed_request_origins)
-          allowed_origins.any? { |allowed_origin| allowed_origin.is_a?(Regexp) ? allowed_origin =~ origin : allowed_origin == origin }
-        end
-
         def respond_to_successful_request
           websocket.rack_response
         end
--- a/lib/action_cable/connection/base.rb
+++ b/lib/action_cable/connection/base.rb
@@ -172,12 +172,17 @@ module ActionCable
         def allow_request_origin?
           return true if server.config.disable_request_forgery_protection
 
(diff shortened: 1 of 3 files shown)


lib/action_cable/connection/base.rb
```diff
@@ -180,6 +180,11 @@ def allow_request_origin?
           end
         end
 
+        def allowed_origins_match? origin
+          allowed_origins = Array(server.config.allowed_request_origins)
+          allowed_origins.any? { |allowed_origin| allowed_origin.is_a?(Regexp) ? allowed_origin =~ origin : allowed_origin == origin }
```
lsylvester: You could use the `===` operator to match both strings and regexes, so the class check wouldn't be required.

```
allowed_origins.any? { |allowed_origin| allowed_origin === origin }
```

lib/action_cable/connection/base.rb
```diff
@@ -180,6 +180,11 @@ def allow_request_origin?
           end
         end
 
+        def allowed_origins_match? origin
```
dhh: Style: Use parenthesis in method definition signature.

lib/action_cable/connection/base.rb
dhh: Actually, given this, we don't need the extra method. We can just inline this === check like we had the original include call.

adamliesko: Sure, that makes sense. Thanks.

dhh:
@adamliesko Do you want to wrap up these changes? Then we can merge.

adamliesko:
@dhh I have updated the `README.md` and rebased the branch anything else left to do?

lib/action_cable/connection/base.rb
lsylvester: @adamliesko I don't think this method is used anymore and can be removed.

dhh:
👏

PR #139, opened 2015-12-14
Introduce cable Rack Middleware

kbrock:
This allows a single rails app to support websocket requests

I added the following to the bottom of my `config/initializers/{development,production}.rb`:

```ruby
config.middleware.use "ActionCable::Middleware"
```

/cc @jmbejar and @SophieDeBenedetto Thanks for sharing this information. This ok to put into `actioncable`?

### DIFF
--- a/lib/action_cable.rb
+++ b/lib/action_cable.rb
@@ -25,6 +25,7 @@ module ActionCable
     autoload :Connection
     autoload :Channel
     autoload :RemoteConnections
+    autoload :Middleware
   end
 end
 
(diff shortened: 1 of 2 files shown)


jmbejar:
Hi @kbrock. I haven't play with ActionCable for a while but I noticed there is already a recommended way to have the cable server responding in the app server process. It's done by mounting the cable server in the router as it's explained in the Readme: https://github.com/rails/actioncable#in-app

kayakyakr:
To echo @jmbejar a bit, what is the implication of using it as rack middleware. Does this just move responding to websocket calls to the root url?

I believe I would still prefer the route version as it would more easily allow you to transition from in-app to multi-server using nginx location directives. Otherwise, you would have to write the detection of the websocket request into nginx itself.

kbrock:
@jmbejar @kayakyakr thanks so much for this.
I had this example kicking around and didn't re-read the readme.

PR #142, opened 2016-01-01
Remove source code

georgeclaghorn:
To avoid confusion now that Action Cable lives [in Rails proper](https://github.com/rails/rails/tree/master/actioncable), leave only the README behind. Matches [what was done](https://github.com/rails/activejob/pull/109) to rails/activejob post-merge.


rafaelfranca:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

